PR #11, opened 2024-07-05
Full System update feature For LinuxTool Box 

TheGamingWinner:
@ChrisTitusTech I have noticed that one of the necessary feature of a tool box to allow full system updates is missing so I added one through this script, It goes ahead and do a whole system update with the following things also :-

- Git repo updates
- pip updates for python
- distrobox updates

The repo used in this script https://github.com/thegamingwinner/update_script.

### DIFF
--- a/src/commands/rofi-setup.sh
+++ b/src/commands/rofi-setup.sh
@@ -1,0 +1,87 @@
+#!/bin/bash
+
+RC='\e[0m'
+RED='\e[31m'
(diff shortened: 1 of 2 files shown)


TheGamingWinner:
Hey @ChrisTitusTech , I also want to know if you want to have reference script like these (that run off other repos) or do you want to include the scripts with the project, Itself

I have scripts that would setup the base for any coder / programmer but need some clarification

PR #12, opened 2024-07-07
Save binary as a tmp file

Pigges:
Instead of creating a directory in home, I've changed the script to save it as a tmp file and deletes it from your system once you are done with it.
I've also changed it to use sh instead of bash, because it runs a bit faster (provided you have linked /bin/sh to dash or something similar, instead of bash).

The startup command is changed to pipe the curl output into sh instead because it's a bit cleaner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,5 +3,5 @@
 ## Usage
 Open your terminal and paste this command
 ```bash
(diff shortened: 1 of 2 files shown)


Pigges:
Test it:
```bash
curl -fsSL https://raw.githubusercontent.com/Pigges/linutil/startup/start.sh | sh
```

ChrisTitusTech:
Very clean! Thanks @Pigges

PR #13, opened 2024-07-11
Added passthrough to the terminal child process

JustLinuxUser:
- Moved the key handling of the running command inside the RunningCommand struct
- Made so that the running command passes through all key inputs to the virtual terminal, except <Ctrl-c> which works as before, by sending SIGHUP to the child process
- Added a useful example "Full bash" and "Full zsh"


ChrisTitusTech (approved):
Neat! I'm going to try this on stream.

PR #30, opened 2024-07-15
fix things show by `cargo check` and `cargo clippy`

Jak2k:


ChrisTitusTech:
Thanks for the suggestions @Jak2k I implemented these but declined the PR so I can use the fail_color inside the program. It wasn't utilized, but I fixed that.

PR #32, opened 2024-07-15
Changed scripts to bash (closes #29) AND removed duplicate code (closes #28)

### LINKED ISSUE #28: Source `common-script.sh` in the other script instead of duplicating

### LINKED ISSUE #29: Bin/sh is stupid here is why
/bin/sh is the default shell. And ChrisTitus thought it was faster. It can technically be. But in Titus'es case it was not. 

In Arch bin/sh symlinked to bash. So its not actually faster. 
In Debian its symlinked to Dash which is faster than bash. 
This is why Titus thought bin/sh was faster... its not.

And if somebody tries to make their favourite shell fish. Making bin/sh fish will make it not work very well.

afonsofrancof:
This PR closes #29. As stated by Meliketoaste /bin/sh is linked to the default system shell, and can be any shell, not just POSIX ones.
Check the comment bellow that explains the other changes

### DIFF
--- a/src/commands/common-script.sh
+++ b/src/commands/common-script.sh
@@ -6,23 +6,26 @@ YELLOW='\033[33m'
 GREEN='\033[32m'
 
 command_exists() {
-    which $1 >/dev/null 2>&1
+    which "$1" >/dev/null 2>&1
 }
 
-checkEnv() {
+
+checkCommandRequirements() {
     ## Check for requirements.
-    REQUIREMENTS='curl groups sudo'
+    REQUIREMENTS=$1
     for req in ${REQUIREMENTS}; do
-        if ! command_exists ${req}; then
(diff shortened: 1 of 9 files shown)


afonsofrancof:
I added more code on the main branch, which was automatically added to this PR.
I wanted to do it on another branch so that I could create a new PR.

The code I added closes #28 . It makes all code depend on common-script.sh , by having the repetitive code there.
checkEnv has been split into multiple functions which are then called by the other files when needed.

Jak2k (approved):

ChrisTitusTech:
Sadly you can't add or source other bash files from inside the linutil. We are working on a common library that gets passed to the script. I also want to stay away from using bash right now and might do some error checking while specifying sh or checking POSIX compliance before launch.

PR #33, opened 2024-07-15
Fix colors with echo commands

luca0N:
Use the `-e` option when using the `echo` command, as backslashes are not interpreted by default.  This causes `echo` to print color escape codes as-is.  This PR also replaces existing `printf` calls with `echo` for consistency.


ChrisTitusTech (approved):
Thanks for these additions. I'd say lets standardize things and any inconsistencies can be a separate PR.

PR #38, opened 2024-07-16
Add Void Linux XBPS support for Full System Update

NightmistTheCat:


ChrisTitusTech (approved):
looks good! I still need to try out void one day...

PR #39, opened 2024-07-16
Formatting Fixes

adamperkowski:
Did a few little formatting changes. 
There is something like [Rust FMT](https://github.com/rust-lang/rustfmt) (an official tool, bundled in cargo) which helps you keep your code "clean". You can use `cargo fmt --all --check` to check your code formatting.
You can also add a GitHub Action that checks it (like [this one](https://github.com/adamperkowski/highlightos/blob/e64bd335cd379712576e1295218d3f05f7694fd8/.github/workflows/rust.yml#L28)). Really helps.


adamperkowski:
Also, you shouldn't really have dead code in your releases. I understand there's dead code when you're actively working on something, but you can always use `debug_assertions` or just comment it out so it won't be there when you run `cargo build --release` and release it.

adamperkowski:
You can also use ex. `cargo clippy` for more lints (which returns one warning btw) but I'm leaving it out to you.

adamperkowski:
This is not a conflict, this is a part of my PR. I have no way of resolving that.

ChrisTitusTech (approved):

ChrisTitusTech:
Thanks @adamperkowski I'm still looking through the debug_assertions you added and seeing a few build errors. I made a couple commits trying to clear this up, but they still remain. I'll checkout this PR and see if I can troubleshoot further tomorrow.

ChrisTitusTech (changes requested):
fix build errors

adamperkowski:
Sure. I'll update the branch and fix everything in about 5h.

PR #41, opened 2024-07-16
A bunch of improvements to the project

AlbydST:
### General Improvements

- Added support for doas and regular su to all scripts
- Removed sudo from requirements everywhere
- Fixed spelling of AUR
- Updated Linutil binary

### Removal of Obsolete Packages

I added a script that automatically removes Obsolete Packages (orphans) from your system.
Currently, it supports:
- Debian (APT)
- Fedora (DNF)
- Arch (Pacman)
- OpenSUSE (zypper)
- Gentoo (Portage)
- Void (XBPS)
- NixOS and other systems with Nix
- Slackware (SlackPKG)
- Alpine (APK)

### Project README

I added a bunch of userful information to the Project's README, such as:

- Linutil's current features
- Currently Supported Systems
- How to contribute to the project

I also added myself to the credits ;)

### List.rs

In addition to adding the obsolete package remover to list.rs, I also made it so that the scripts that are cURLed from the web (e.g. mybash) are piped to `sh` instead of being ran with `bash -c`. This makes it so that they don't rely on `bash`.

### System Update Script

Added support for:
- Gentoo (Portage)
- Void (XBPS)
- NixOS and other systems with Nix
- Slackware (SlackPKG)
- Alpine (APK)

### Common Script

Added other package managers as an example:
- XBPS
- Nix
- SlackPKG
- APK
- Eopkg
- YUM
- Rad
- KISS


ChrisTitusTech:
Love a lot of the ideas but these really need to be seperate PRs. there are conflicts with some of the updates so it is best to break out each big fix into it's own pr. Ex. Orphan PR, list.rs fixes, change sudo to SUDOCMD.

PR #42, opened 2024-07-16
minor: shorter oneliner

### LINKED ISSUE #22: Prettier one-liner
This the the current one-liner:

```bash
curl -fsSL https://github.com/ChrisTitusTech/linutil/releases/latest/download/start.sh | sh
```

But this one looks better :D

```bash
bash <(wget -qO- https://christitus.com/tux)
```

thecmdrunner:
This the the current one-liner:

```bash
curl -fsSL https://github.com/ChrisTitusTech/linutil/releases/latest/download/start.sh | sh
```

But this one looks better :D

```bash
bash <(wget -qO- https://christitus.com/linux)
```

I know the URL does not point to the script yet, but would be awesome to see this one go live!

closes #22

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 Open your terminal and paste this command
 ```bash
-curl -fsSL https://github.com/ChrisTitusTech/linutil/releases/latest/download/start.sh | sh
+bash <(wget -qO- https://christitus.com/linux)
 ```
 
 ## Credits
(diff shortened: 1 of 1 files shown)


thecmdrunner:
Makes sense. `curl` isn't installed on Ubuntu on default tho :)

ChrisTitusTech (approved):
Ended up with curl as it is the de facto standard for transfers as MacOS / Windows has it loaded by default where wget gets left out. Most linux distros come with curl and NOT wget. So curl will win this battle.

PR #43, opened 2024-07-16
Add bitmap font rendering toggle feature

taovuokko:
I consistently encounter issues where text in Firefox and Thunderbird appears thin, ragged, and hard to read, especially with fonts like Calibri. The issue persists even after installing the required fonts. After some research, I found that the solution is to disable bitmap font display.

This pull request adds a simple toggle for bitmap font rendering, which can significantly improve text appearance.

**What I added:**

- New script: `3-toggle-bitmap-font-rendering.sh` in `commands/system-setup/` 
- Updated `list.rs` to include the new option under "System Setup"

**How it works:**
The script creates or removes a fontconfig file that disables embedded bitmaps for all fonts. Users can easily toggle this feature on/off through the CLI menu, with clear instructions provided.

This small change can make a big difference in readability without manual config edits.

Inspired by: https://wiki.archlinux.org/title/font_configuration#Disable_bitmap_fonts


philipprochazka:
I think I like it ! But I havent checked the code 😸

taovuokko:
> I think I like it ! But I havent checked the code 😸

Thanks! I noticed after submitting that other scripts use #!/bin/sh -e. I used #!/bin/bash as this script uses Bash-specific feature. Should I change it for consistency?

ChrisTitusTech (approved):
Very interesting, I never knew this was a thing I'll test this out.

taovuokko:
> Very interesting, I never knew this was a thing I'll test this out.

Issue primarily affects older Microsoft Office fonts. The rendering engine typically uses embedded bitmap images in fonts, designed for Windows rendering. However, this approach doesn't work well with most Linux rendering engines. This configuration disables bitmap usage, forcing vector rendering instead.

I've personally had to face this problem numerous times, as my day job requires use of Calibri font. Even slight size reductions make it blurry and hard to read on Linux systems

ChrisTitusTech:

PR #48, opened 2024-07-17
Add easy way to do local development and move checkEnv to the start scripts

Subarctic2796:
Hi Chris.

I have made 2 main changes.

## Add an easy way to do local development
I updated the README to include instructions for how to contribute.
Basically I added a variable called `ISLOCAL` to the `startdev.sh` script and if you want to do local development you change the `ISLOCAL` variable from `false` to `true`. So instead of calling down the binary from the github release to building and running the local binary and the local scripts. So it will always use the local changes that you added.

## Move `checkEnv` function from each script to the `start.sh` script
I have moved the `checkEnv` function from each script and moved it to the start scripts. This way you aren't running it every single time you run one of the scripts from the util. This will make it faster.


ChrisTitusTech (approved):
Had to eyeball these conflict merges on a browser, but need testing after I modified code contributed from other PRs.

ChrisTitusTech:
Ended up going with a concatenation in the rust file with common_script getting added to the header of all scripts run in the commands folder.

Subarctic2796:
I understand, but I have a few issues with how this is done.

1. It inflates the final binary size as every new script is compiled into the binary, and now every script is going to be even bigger making the final binary also much bigger.
2. Every script is running the common script every time which is redundant.

Whereas the way I was doing it you only had to run the common script once to get environment variables. And also doesn't need to be included in every script.

You don't read to reply, but it would be nice to get a reply.

PR #60, opened 2024-07-21
refactor: Use common script to avoid code duplication

### LINKED ISSUE #28: Source `common-script.sh` in the other script instead of duplicating

lj3954:
This proposed change includes creating a macro to concatenate the common script with the specified script. A macro must be used, since macros (including the concat macro) are evaluated before functions.

The shell will be launched through the shebang within the common script, shebangs within other scripts will be disregarded as comments when they reach the interpreter. This wouldn't cause issues with these scripts, but it should be noted in case the arguments to the shell need to be modified in the future.

Closes #28


afonsofrancof:
You might not have realized that different scripts have slightly different checkEnv functions.

I submitted a PR #32  a few days ago that correctly handled this behavior and even separated checkEnv into smaller functions.
 It was rejected since I was sourcing the common script inside the shell scripts, instead of concatenating, like you are doing. 
I think the best way to solve this is for you merge your changes to the rust code with mine to the shell scripts.

You can pull my PR to inspect it by cloning this repo (It's the one that stores the PRs) and pulling the PR 

```bash
git clone https://github.com/ChrisTitusTech/linutil

git fetch origin pull/32/head:pr_32

git switch pr_32
```

alternatively if you have this repo as upstream in your fork's clone , just do 

```bash

git fetch upstream pull/32/head:pr_32

git switch pr_32
```

afonsofrancof:
Nice! I think you might need to update your branch to be compatible with the new main now.
They have changed system-update.sh, and now supports "xbps-install", so it requires it to be added when checking the package manager, as well as to the fastUpdate function.

lj3954:
That should be all that's needed. No changes are necessary to the fastUpdate function.

afonsofrancof:
> That should be all that's needed. No changes are necessary to the fastUpdate function.

Yup, my bad.
Just tested your PR and everything is working perfectly.

ChrisTitusTech (approved):

PR #69, opened 2024-07-25
nixos support for dotfiles + graphical exit

guruswarupa:
# Pull Request

## Changes made
-all the dotfiles
-list.rs


ChrisTitusTech:
moved a lot of this into the common script that now get concatenated with every sh script run.

PR #75, opened 2024-07-25
Add ZSH install option

jtmusselman:
# Pull Request

## Title
I wish to add ZSH and a simple prompt

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation Update (Due to be added!)
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
I added a script to add zsh, and some fixes to move .zshrc to ~/.config/zsh/.zshrc

## Testing
Ran the script on a newish arch machine without zsh

## Impact
Allows users

## Additional Information
Probably should be tested on more machines. First time contributing to a project that is not my own.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project. (I think)
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #82, opened 2024-07-27
Add `. filename`  support

JustLinuxUser:
# Pull Request
## Add support for sourcing sh files (or any runtime filesystem based operation)
I added a crate called `include_dir`, which will compile in a whole directory, and when linutil is started, it will unpack that directory to a temp directory. And because now we are working with a full filesystem, no hacks with concatenating files are needed.

I also added a struct `AppState` that can be passed around and hold preferences that need to be accessed all around the app, like theme, and the path to the unpacked files

## Why concatenation hacks are bad?
- The errors inside the scripts aren't useful anymore, because the line numbers are broken
- We are limited to a couple of libraries before things get out of hand
- It's harder to test/use the scripts outside `linutil`

## Type of Change
- [x] New feature
- [ ] Bug fix
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Additional Information
[Any additional information that reviewers should be aware of.]

## Checklist


lj3954:
The source command cannot be used, it's not included in the posix spec so it's not guaranteed the shell /bin/sh is symlinked to will support it (Debian uses Dash, for example, which doesn't support it). I also highly doubt you've tested much of this, since most of the scripts simply don't work. You're incorrectly trying to source the common script from the script's directory, rather than the script's parent directory, in multiple scripts. Simple testing would reveal that the path doesn't exist.

I oppose this change in general, I disagree that more external dependencies should be added and that temp directories should be created for such minor grievances. Line numbers? It's not too difficult to subtract the length of the common script (85), and steps which fail usually are commands which only appear on one line regardless.

If it's desired to use source to improve local development and usage outside of linutil, I would propose instead adding a build script which recursively iterates over the files within the src/commands/ directory and replaces `source $file` or `. $file` with a copy of that file. This would address all of your concerns other than line numbers.

One commit also probably also shouldn't have multiple unrelated changes with a 400+ diff.

JustLinuxUser:
@lj3954 I agree, my MR is a mess, and I didn't know that source is not POSIX complient, but if `. file` is, it's fine.
Sourcing from any directory is fine, because of how this scripts are run, they are always launched from `commands` directory (or unpacked version in /tmp), so that is not a problem.
I have thought about a build step, but I would like to avoid adding that at all cost, and if you have a full fs to work with, you can add more then just sourcing multiple files, you can have assets that are copied in that don't deserve their own git repo, you can have small configs inside, and in general it's much easier to work with a full fs.

Local testing and easy work with local files is also very good, it allows to use normal IDEs and LSPs for autocomplete, and again if a build step is not needed, It should not be added

JustLinuxUser:
I want to add that out of the 400 changed lines, 
80 is Cargo.lock
and ~200 is me switching everything to use AppState, so it's not actually that big of a change

JustLinuxUser:
I changed all `source` statements to `. filename` btw, so this should work on debian too,  I tested it in distrobox briefly

lj3954:
> can add more then just sourcing multiple files, you can have assets that are copied in that don't deserve their own git repo

I cannot think of a use case where bash scripts used to setup a system would depend on any file other than a different bash script. Any assets used within the Rust program, rather than in scripts, should never be stored in this strange temporary directory. They should be include!()'d directly. In fact, that's another great argument for modification at compile time. Duplicate assets would be created if this temporary directory included anything which the Rust program relied upon.

> I have thought about a build step, but I would like to avoid adding that at all cost

I don't see a reason why a build step should be avoided. Modifying text files takes an insignificant amount of time. The only disadvantage I can see is that your issue regarding line numbers remains. Rust includes tooling to create build scripts which automatically are called by cargo, nothing extraordinary needs to be done.

> and ~200 is me switching everything to use AppState, so it's not actually that big of a change

My point was that the AppState changes are clearly unrelated to your temporary directory. These should not be part of the same commit.

JustLinuxUser:
>I cannot think of a use case where bash scripts used to setup a system would depend on any file other than a different bash script. Any assets used within the Rust program, rather than in scripts, should never be stored in this strange temporary directory. They should be include!()'d directly. In fact, that's another great argument for modification at compile time. Duplicate assets would be created if this temporary directory included anything which the Rust program relied upon.
Yes, bash files don't "depend" on anything besides other bash files, but if you have access to source (`. filename`), it's much much easier to add a couple of dependencies. As for assets, Imagine a directory inside `commands` called `assets` or something, and then for the script to use any of those assets it can just `cp assets/pfp.jpg new-location` I don't see how that would create a duplicate in any way shape or form, and I am not talking just about pictures, small configs, etc. It is just easier to work with.

>I don't see a reason why a build step should be avoided. Modifying text files takes an insignificant amount of time. The only disadvantage I can see is that your issue regarding line numbers remains. Rust includes tooling to create build scripts which automatically are called by cargo, nothing extraordinary needs to be done.
A build step requires an additional command + dependency to be executed before the actuall compile step, and if there is no reason to have one, why?

>My point was that the AppState changes are clearly unrelated to your temporary directory. These should not be part of the same commit.
As I was writing my code I realized AppState would be nice, so I just started adding it at the same time, But I really see your point, and I will split that commit later today

lj3954:
> Yes, bash files don't "depend" on anything besides other bash files, but if you have access to source (`. filename`), it's much much easier to add a couple of dependencies

My PR specifically solves this, by iterating through all scripts, replacing `source $file` or `. $file` with a copy of that file, right before the program is compiled. This also erases the possibility of _**any**_ issue with source at runtime, since if the file being added doesn't exist, linutil won't compile. 

> It is just easier to work with

Yes, it's easier to copy images if you have a temporary directory. But there's no reason linutil would ever have to copy images around. I think it's very unlikely any shell script, anywhere, but especially for linutil, would need these other non-shell files. Assets which linutil, the rust-based TUI, relies on, would be include!()'d anyways, as already mentioned. That's where I would expect images to be used, but a temporary directory is entirely unnecessary for that. 

> A build step requires an additional command + dependency

First, the tooling is fully automated with cargo. Just adding a build script (build.rs) does all that's needed. The build script is also not re-run unless x happens that you specified in the previous run of the build script (i.e. a script or the build script is modified). Creating a temporary directory **at runtime** is also significantly more impactful than slight modifications to text files at compile time. I've done a few further optimizations to ensure there's essentially zero hit to compilation times. The file extension will be validated first (none or .sh), and then only the first 2 bytes (rather than the entire file) are read in order to find a shebang. No time will be wasted in the case of large files.

JustLinuxUser:
You have 2 options:
- Have a compile time pre processor that replaces all `source x` and `. x` with it's file, needs a weird macro, makes it hard to debug, and creates an additional build step, makes it hard to manage assets like small configs
- Just have a complied in directory.
I feel like the choice is kind of obvious.
I am not trying to attack you, and I know that it is not fun to have your code ignored / removed. But sometimes simplicity is just better, and it would be nice if you admitted that. Your solution was the first thing I thought about, and creating a build step is a natural conclusion to jump to. But look at it this way:
To make this possible you need to add:
- build.rs
- an additional macro
- A directory with built files (which doubles the commit size for any script commit unless you explicitly ignore it)
- A pre processor step that is prone to breaking (imagine it's a part of a multi line string with `<<<EOF` or a multi line comment. or an additional space before `source` or in between the source and the filename

JustLinuxUser:
I split my commits to improve the quality of the commit messages, and reduce their scope

lj3954:
> simplicity is just better

This is the same 'simplicity' people are extremely worried about while writing JavaScript for something that JavaScript definitely shouldn't be used for. Pulling in every package from who knows where as a dependency affords you nothing but uncertainty. 

The build script I wrote is rather simple & well commented. The macro is also very simple and does nothing but redirect `src/` to `${BUILD_DIR}/src/` in the call to include_str. Nobody checks the build directory into version control. If you do that, you have much more to worry about than <1KB shell scripts, since it's full of binaries. 

I think I've addressed well enough the problem of multiline strings.

JustLinuxUser:
Put it this way:
What can your PR do that mine can't? - Nothing really
But with this PR you have:
- Much less maintenance (everybody knows all the limitations of directories)
- The ability to include arbitrary files inside the temp directory (Which is very useful for configs, and can even be implemented as a git submodule)
- Running a script manually will do the exact same thing and have the exact same warnings and errors as it would when running inside linutil. (No multiline strings, no spaces braking scripts, no crashing the PC on the compile time)
- Simplicity of use from the DX perspective, no macros, no build step to think about.

lj3954:
> Running a script manually will do the exact same thing and have the exact same warnings and errors as it would when running inside linutil. (No multiline strings, no spaces braking scripts, no crashing the PC on the compile time)

Running a script in which way? You'll have to be in the src/commands directory, no matter which directory the script is in. Most people locally launch scripts from the directory they're located in, not the parent directory of that. 

Again, determining errors at compile time is a better solution than at runtime. That's a very clear example of something my solution does that yours doesn't. The issue you brought up with spaces is fixed now, although the linter shouldn't allow that in shell scripts anyways. 

> Which is very useful for configs

Configs are usually important enough to belong on a different repository. Submodules are a great idea for configs and other constantly changing files if you love constantly updating them. Small configs, which don't need their own repository, are often left within the script as a multiline string anyways. PR #75 already proposes doing that in one case.

> no macros, no build step to think about

A very simple macro doesn't hurt anyone. You don't need to think about the build step any more than you'd have to think about your temporary directory creation.

lj3954:
`sh $file` cannot be used for any script which relies on a different shell. The contents of the shebang should be respected.

JustLinuxUser:
@ChrisTitusTech Do you think this PR needs something added / changed? If so I have some free time, so it would be no problem for me to do

ChrisTitusTech (approved):

PR #83, opened 2024-07-27
chore: Build linutil against musl libc

### LINKED ISSUE #37: GLIBC Version Error When Running Script on Windows WSL 2.1.5.0
## Environment
- **Operating System**: Windows Subsystem for Linux (WSL) 2.1.5.0
- **Distribution**: Debian

## Issue Description
I am experiencing a version mismatch error related to GLIBC when attempting to run a script on Windows WSL. Below are the error messages that are displayed when I try to execute the script:
```sh
/tmp/tmp.w9U597xQHU: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/tmp.w9U597xQHU) 
/tmp/tmp.w9U597xQHU: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/tmp.w9U597xQHU)
/tmp/tmp.w9U597xQHU: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/tmp.w9U597xQHU)
ERROR: Executing linutil
```

lj3954:
# Pull Request

## Title
Statically link linutil binary against musl libc 

## Type of Change
- [x] Bug fix

## Description
Modify GitHub workflows to build the linutil binary against the x86_64-unknown-linux-musl target. This leads to the binary being statically linked against the musl libc, removing the glibc dependency. This will improve compatibility on distros which include a different libc or a build of glibc which isn't supported by rust's x86_64-unknown-linux-gnu target in the current toolchain version. 

## Testing
The TUI successfully builds against the musl target, and successfully launches.

## Impact
The size of the binary increases by a small amount (~100KB) due to static linking. The performance impact should be negligible.

## Issue related to PR
- Resolves #37

## Additional Information
Other changes to the workflow files are from format on save

## Checklist

### DIFF
--- a/.github/workflows/linutil.yml
+++ b/.github/workflows/linutil.yml
@@ -2,7 +2,7 @@ name: LinUtil Release
 
 on:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
 
 permissions:
   contents: write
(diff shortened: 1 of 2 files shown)


afonsofrancof:
I think we should maybe move the file from `build/x86_64-unknown-linux-gnu/release` back into `build/release` before auto-committing, so the directory structure is cleaner.

lj3954:
I'll leave that decision to Chris. He can push a commit here if he'd like to make such a change. You could make the argument for moving it anywhere. Perhaps just build/, or something else like the root of the repository.

lj3954:
This won't have an affect on aarch64, a separate build would be required for those (aarch64-unknown-linux-gnu/musl). This only fixes compatibility with non-glibc distros or linux systems with an incorrectly built or incompatible glibc version, as in #37.

ChrisTitusTech (approved):

PR #85, opened 2024-07-28
refactor: Handle `source` and `.` in scripts during build

lj3954:
# Pull Request

## Title
Handle `source` and `.` invocations in scripts during the build process.

## Type of Change
- [x] New feature
- [x] Refactoring

## Description
Remove macro to concatenate common-script with the given script. Instead, create a build script which recursively iterates through the src/commands/ directory and gathers a list of scripts (identified based on the beginning shebang). Each script is then modified (only in the build directory, not the actual repository) to replace lines of `source $file` or `. $file` with a copy of that file. A macro is then added to list.rs (include_script) in order to source the modified scripts from the build directory, rather than the unmodified scripts from the repository.

#82 attempts to implement this same feature, but it goes about it in a different way. Instead of modifying the files in the build step which will reliably include a copy of the script in the program, it uses 2 extra crates, one of which being unmaintained for the last 6 years, to include an entire directory in the build process, and create a temporary directory including those files at runtime.

- Closes #82 

## Testing
I've tested multiple scripts, including those in directories other than the same directory as common script. 

## Impact
Scripts can be used outside of linutil, and any file in the repository can be sourced if necessary, rather than just the common script. Scripts are also launched with the shebang contained within them, rather than the one in common-script, so if some script requires bash or another shell rather than a generic posix shell, it would be fully respected.

## Checklist

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -55,7 +55,7 @@ fn replace_source(file: &Path) -> String {
         .map(|line| {
             if line.starts_with(". ") || line.starts_with("source ") {
                 let (_, sourced_file) = line.split_once(' ').unwrap();
-                let sourced_file = filedir.join(sourced_file);
+                let sourced_file = filedir.join(sourced_file.trim_start());
                 if !(sourced_file.exists() && has_shell_ext(&sourced_file)) {
                     return line.to_string();
                 }
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,65 @@
+use std::{
+    env, fs,
+    io::Write,
+    path::{Path, PathBuf},
+};
+
+const SCRIPT_PATH: &str = "src/commands/";
+
+fn main() {
+    let out_dir = env::var("OUT_DIR").unwrap();
+    let file_list = get_script_list(Path::new(SCRIPT_PATH));
+    replace_source(file_list, out_dir.into());
+}
+
+fn get_script_list(path: &Path) -> Vec<PathBuf> {
+    let paths = path.read_dir().unwrap();
+    paths
+        .into_iter()
+        .flatten()
+        .filter_map(|entry| {
+            let path = entry.path();
+            // Recursively iterate through directories
+            if entry.file_type().map_or(false, |f| f.is_dir()) {
+                get_script_list(&path).into()
+            } else {
+                let is_script = fs::read_to_string(&path).map_or(false, |f| f.starts_with("#!"));
+                is_script.then_some(vec![path])
+            }
+        })
+        .flatten()
+        .collect()
+}
+
+fn replace_source(files: Vec<PathBuf>, out_dir: PathBuf) {
+    for file in files {
+        println!("cargo:rerun-if-changed={}", file.display());
+
+        let mut out_file = create_out_file(&file, out_dir.clone());
+        let contents = fs::read_to_string(&file).unwrap();
+        let filedir = file.parent().unwrap();
+
+        let new_file = contents
+            .lines()
+            .map(|line| {
+                if line.starts_with(". ") || line.starts_with("source ") {
+                    let (_, sourced_file) = line.split_once(' ').unwrap();
+                    let sourced_file = filedir.join(sourced_file);
+                    std::fs::read_to_string(&sourced_file).unwrap()
+                } else {
+                    line.to_string()
+                }
+            })
+            .collect::<Vec<_>>()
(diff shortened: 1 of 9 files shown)


JustLinuxUser:
This is a cool idea, but VERY fragile!
1. Doesn't handle multi line comments / strings, so might accidentally include files not meant to be included
2. Handles multiple imports very poorly. Because it only goes 1 level into the dependency chain, and the order matters. For example

```bash
# test1.sh
. test2.sh
```
```bash
# test2.sh
. test3.sh
```
```bash
# test3.sh
echo HI
```
And with a proper bundler the resulting `test1.sh` should be:
```bash
#test1.sh
#test2.sh
#test3.sh
echo HI
```
but because this implementation does not have a notion of a dependency graph, the actual result is:
```bash
#test1.sh
#test2.sh
. test3.sh
```
which would produce a runtime error.

This is far from all issues with this PR, but I think this is enough to show the benefits of pr #82

lj3954:
Your first concern would be difficult to fully address without writing essentially what is a posix interpreter, but https://github.com/ChrisTitusTech/linutil/pull/85/commits/771f06f1275c3c9c45234dbfaf04d08eb06704d5 should somewhat handle it by skipping sourcing rather than panicking on file read failure. 

Your second concern has been fully addressed.

JustLinuxUser:
Again by fixing 1 problem you created another.
1. Your solution will create a lot of code duplication increasing the size of the binary
2. In case of circular dependencies this will use up all the memory on the PC, crashing it in most cases.

lj3954:
> 1. Your solution will create a lot of code duplication increasing the size of the binary

Increasing the size of a binary a few megabytes large, by possibly a kilobyte. How much are your added dependencies increasing the binary size? Of course, you didn't check. 

> 2. In case of circular dependencies this will use up all the memory

As opposed to finding this issue at runtime? Again, that's not a disadvantage. Finding errors at compile time is always preferred to errors at runtime.

JustLinuxUser:
Did you just say that crashing your computer or relying on oom_killer is better then the intended bash behavior? That is not compile time errors, that is a nuke that should not behave like that :rofl:! Imagine Chris is streaming and he accidentally does a circular dep, so now oom_killer will just kill a bunch of random processes and create a bunch of issues for no reason

lj3954:
I figured 'crashing it' was related to the compilation, not actually running out of memory. Here's an actual example of what infinite recursion results in. 

```  thread 'main' has overflowed its stack
  fatal runtime error: stack overflow
```

JustLinuxUser:
Well kind of, it is a race condition between OOM and stack overflow. when the files are small for testing stack overflow will happen first, but for big files, and systems with lower ram, OOM will happen first

JustLinuxUser:
I just tried that, 2 files both of which have 200 loc filled with `f`s ~80 per line, invoke the OOM killer on my 32 gig of ram machine

lj3954:
I assume you put the call to source very late in those files, since that determines how much RAM is used. I added source to the top of the scripts, like you usually would, and the build took about 90MB of RAM before the stack overflowed. I'll fix it regardless.

JustLinuxUser:
Actually it was a stack overflow, I just misinterpreted the crash message

JustLinuxUser:
the actuall OOM happens much much later when the files are about 50 000 loc, so yea, I was definitely wrong on that one

JustLinuxUser:
You did it wrong. 
image
How you implemented that check, it would trigger in the first case, when it should not.

btw. See how this is more maintenance? If I didn't tell you about the first issue it would have been completely broken

JustLinuxUser:
The arrow means sourcing, the node the arrow is being pointed too is sourced by the node the arrow points from.
Your code will brake in a case *like* the first one, but when it gets a bit larger, in your case the limit is 5. And only 3 is drawn on the picture

lj3954:
> but when it gets a bit larger, in your case the limit is 5

My goal was to effectively reduce the size of the stack, not to detect circular dependencies. That would require passing a copy of the calling path as well and require a significant amount of unnecessary logic, since if a shell script is sourcing many other shell scripts, that code is definitely not being written in the correct language. 
 
Regardless, I think with all of these different cases, it's not worth continuing with this approach. I dislike the creation of what are clearly unnecessary temporary directories at runtime, but it'll have to do for now until a better solution is found.

Issue #84, opened 2024-07-27
UI/UX - Tabular model


SoapyDev:
## Some user stories: 
-----------
### User
- As a user navigating the application for the first time, I would like to have a concise view for every aspect that the application can affect on my machine, so I can navigate the application with ease.

### Dev
- As a developer, I would like to be able to add new commands, install application with ease so I can participate.
- As a developer, I would like to be able to participate, without having to understand the whole application, to limit the time spent on adding simple stuff.

## Propositions

The application could use multiple tabular views to segregate abilities, show a welcome message and a help section.
This would allow for more clarity for the end user, while also allowing the code to be easily separated by context.

The idea is to separate every context and its commands to their own domain. That way, every tab can become their own crate that is responsible for rendering, the list of commands, description, names, icons. 

Visual example  (dont pay attention to the colors, I made this with defaults): 
![image](https://github.com/user-attachments/assets/23b87d48-fa91-4d76-a3fb-ce98be788e1a)

bungadrum:
This looks amazing forget about it being just a model or feature/option this has potential to be a new tiling WM or even a new Desktop enviroment. Anyone who likes working with tabs is gonna love this.

JustLinuxUser:
It looks good, but what if the number of tabs is just too big though, and this also makes nesting tabs kinda impossible, rn you can nest directories as deep as you want, but nesting and managing tabs would be much harder. But if Titus likes it I will implement that no questions asked :)

JustLinuxUser:
Another concern I see, is it would be hard to handle resizing terminal to be much smaller, so that would seriously reduce compatibility. But this can be solved by writing the tab widget from scratch or using some widget library for rattatui

JustLinuxUser:
To answer your concern about easily adding new commands as a developer. The file `list.rs` has a variable called [tree](https://github.com/ChrisTitusTech/linutil/blob/main/src/list.rs#L61). When you look at it a little it is very easy to add new stuff to it, it uses a JSONish syntax, so you can easily make PRs by changing that variable. If you need further help after looking at it, you tag me in this issue again and I will write a detailed explination on how to make changes to that variable. Good luck :) and I am impressed that you managed to create that interface as a prototype

SoapyDev:
Hello, 

Thanks for the feedback @bungadrum , @JustLinuxUser . 

- For the first concern, its real, but I would suggest that within a tab, you can still create a nesting list or subcategories. 
[...]

PR #89, opened 2024-07-29
UI/UX - Tabular Prototype

SoapyDev:
# Pull Request

## Title
Tabular Prototype - UI/UX 

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description

Good day,

Here is a prototype for a tabular navigation to simplify the user experience. This prototype aim to flatten nesting structures in a scalable and intuitive way.

Addition : 
Tabs, CommandList, lots of navigation.

### Benefits 

- We can now have a welcome and help tabs dedicated to give direction and guidance to first time Linux users. 
- We avoid nesting structure that can gain complexity as time goes.
- We have a more winutil feel, that is already known and appreciated.
- We intuitively know where to put new scripts.

An emerging benefit is the separation of concerns. Application logic is separated from the view, easing new dev/sys admin that might not understand Rust into a simple and limited amount of code. 

### Inconvenient

- Need to re-implement some previously created feature .

I did try to also re-add the terminal preview, but hats off to the initial creator, its over my understanding of the current crates used by the project.  

## Testing

Only manual testing was done. 

## Impact

Added strum to the Toml as a dependency. This is also widely used by the Ratatui crate in their example.
All the previous mentioned change.

## Issue related to PR
- Resolves # https://github.com/ChrisTitusTech/linutil/issues/84

## Additional Information

Example : 

https://github.com/user-attachments/assets/3cf17eb4-4d45-4f3f-8005-542f22a65ea6

Feel free to chip, remove, take and copy/pasta as you like. As mentioned earlier, this is a prototype. I do not expect this to survive till the end of times! Style and content is only there to show how and where to add it.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,9 +5,9 @@
 
 ![Preview](docs/assets/preview.png)
 
-A distro-agnostic* toolbox which helps with everyday Linux tasks. It can help you set up applications, and set up your system for specific use cases! Written with Rust 🦀
+A distro-agnostic* toolbox which helps with everyday Linux tasks. It can help you set up applications and your system for specific use cases! Written with Rust 🦀
 
-\* — The project is in active development, so there could be some issues, please, consider [submitting feedback](https://github.com/ChrisTitusTech/linutil/issues)
+\* — The project is in active development, so there could be some issues. Please consider [submitting feedback](https://github.com/ChrisTitusTech/linutil/issues).
 
 ## 💡 Usage
 
@@ -20,10 +20,10 @@ curl -fsSL https://christitus.com/linux | sh
 ### [LinUtil Official Documentation](https://christitustech.github.io/linutil/)
 
 ## 💖 Support
-- To morally and mentally support the project, make sure to leave a ⭐️!
+To morally and mentally support the project, make sure to leave a ⭐️!
 
 ## 🏅 Thanks to all Contributors
-Thanks a lot for spending your time helping Linutil grow. Thanks a lot! Keep rocking 🍻.
+Thanks a lot for spending your time helping Linutil grow. Keep rocking 🍻.
 
 [![Contributors](https://contrib.rocks/image?repo=ChrisTitusTech/linutil)](https://github.com/ChrisTitusTech/linutil/graphs/contributors)
 


JustLinuxUser:
> We avoid nesting structure that can gain complexity as time goes.

1. you made it much much more complex on the spot
2. This issue is easily solved by creating subtrees in different files, and then just adding them to the parent node. The only reason this is not implemented in the starting implementation, is that I was not sure this was even necessary.

JustLinuxUser:
Adding a new directory in the current implementation is roughly 5 loc, but with this aproach it seams that amount will grow significantly.

SoapyDev:
As mentioned earlier, this is a prototype. 
Its there to show a different possibility to present things to the end user, with all the complexity, difficulties, and what else that comes with it. This is not expected to replace the current implementation, but open the discussion on an other way to do things.

1. When I spoke about the complexity, its related to how nested structure, tend to become complex as they grow. In this case we have 6 flat structure that are theme based. Scaling is done horizontally, with minimal verticallity. 
2. I agree.


Yes, I did have to rewrite almost all the app, as this is quite a different paradigm  and I could not find a way to simply reuse everything, which I would have loved as I had to write quite a lot. This is after all, my first time using Ratatui, so maybe I just passed next to an easy or not so easy way to reuse what was done. My apologies for that. 

I agree, this implementation incur a lot of boiler plate code. I am sure most of it can be improved, but I was more concerned about having a minimal viable product than reducing LOC. 

As always, thanks for your time and feedback.

JustLinuxUser:
> As mentioned earlier, this is a prototype.
Its there to show a different possibility to present things to the end user, with all the complexity, difficulties, and what else that comes with it. This is not expected to replace the current implementation, but open the discussion on an other way to do things.

Than you did an amazing job! Thanks! I will re read your code tomorrow, and mb try working some of your ideas in the main project

ChrisTitusTech (commented):
Very Cool love to see where this goes.

PR #105, opened 2024-07-30
DWM-Titus fixed and ready

infstate:
Finally fixed DWM-Titus and works with the new UI


infstate:
@ChrisTitusTech I  would say that you should merge this and since its easy to read, modify it yourself to setup DWM-Titus how you would want it on a users PC. Just change or add a few commands and it should be fine.

ChrisTitusTech (approved):

PR #108, opened 2024-07-31
Fixed the wrong command on selection bug in search.

afonsofrancof:
Fixed the bug talked about on stream.

Also, @ChrisTitusTech, you implemented scroll logic, but there is a built-in scroll up and down function in the crate we are using for the list, so I removed your function and added the built-in one.


afonsofrancof:
I was using rust 1.79 , and github actions was using 1.80, that's why my cargo fmt was failing.....

ChrisTitusTech (approved):

ChrisTitusTech:
Looks great, thanks @afonsofrancof

PR #119, opened 2024-07-31
Added missing inclusion of common-script on firewall-baselines

afonsofrancof:
The import of common-script was missing on firewall-baselines


afonsofrancof:
#123 fixes this and some other stuff

PR #120, opened 2024-07-31
Move search bar widget to a separate file

afonsofrancof:
This pull request moves the search bar logic to a different file, for better organization.


src/search.rs
```diff
@@ -0,0 +1,85 @@
...
+        //Set the search bar text (If empty use the placeholder)
+        let display_text = if self.search_input.is_empty() {
+            if self.in_search_mode {
+                Span::raw("")
+            } else {
+                Span::raw("Press / to search")
+            }
+        } else {
+            Span::raw(&self.search_input)
+        };
```
JustLinuxUser, line 47: this could be simplified to:
```rust
let display_text = if !self.in_search_mode && !self.search_input.is_empty() {
     Span::raw("Press / to search")   
 } else {
    Span::raw(&self.search_input)
}
```
Or some inversion of that logic

JustLinuxUser (commented):
Also please squash your commits, if your commits don't have sematic meaning (you did different things with each commit that still makes sense afterwards). An example of good commit splitting is something like:
1. Added a new feature to x
2. Made Y use x
3. Made Z use x
A bad example of how not to do that is:
1. added a new feature
2. Fixed bug 1 in that feature
This doesn't make sense on the commit level, because it can just be squashed together.

In case you don't know how to sqash commits, here is a quick rundown.
1. `git reset --soft <hash before your first commit>`
2. `git commit "New commit message"`
3. `git push --force`

afonsofrancof:
> Also please squash your commits, if your commits don't have sematic meaning (you did different things with each commit that still makes sense afterwards). An example of good commit splitting is something like:
> 
> 1. Added a new feature to x
> 
> 2. Made Y use x
> 
> 3. Made Z use x
> 
> A bad example of how not to do that is:
> 
> 1. added a new feature
> 
> 2. Fixed bug 1 in that feature
> 
> This doesn't make sense on the commit level, because it can just be squashed together.
> 
> 
> 
> In case you don't know how to sqash commits, here is a quick rundown.
> 
> 1. `git reset --soft <hash before your first commit>
> 
> 2. `git commit "New commit message"`
> 
> 3. `git push --force`

Yup, thanks.
I did look into squashing the commits yesterday, but ended up not doing it because it was late.
Will pay more attention to that from now on.

JustLinuxUser:
I accidentally forgot to close the `\`` in my comment, make sure you re read the portion about sqashing the commits, because the first command got cut off. Sorry about that

afonsofrancof:
And once I get home I will squash them

src/search.rs
afonsofrancof: > this could be simplified to:
> 
> ```rust
> let display_text = if !self.in_search_mode && !self.search_input.is_empty() {
>      Span::raw("Press / to search")   
>  } else {
>     Span::raw(&self.search_input)
> }
> ```
> 
> Or some inversion of that logic

There should be no negation on the second part of that if statement, but the rest seems good.
I will commit it and add you as co-author.

JustLinuxUser:
I am Justliknuxuser and I endorse this PR :smile:

ChrisTitusTech (approved):

PR #126, opened 2024-08-03
Introduce Automatic Global Theming and some fixes

### LINKED ISSUE #122: Why creating Linuxtoolboxdir in 3-global-theme.sh
https://github.com/ChrisTitusTech/linutil/blob/6f865f76efd1645b082b0132831f4f05877a8418/src/commands/system-setup/3-global-theme.sh#L6-L14

Why are we creating the linuxtoolbox directory? It has no practical use,Instead of helping, it is causing problems with

https://github.com/ChrisTitusTech/linutil/blob/6f865f76efd1645b082b0132831f4f05877a8418/src/commands/system-setup/3-global-theme.sh#L73

which calls
https://github.com/ChrisTitusTech/linutil/blob/6f865f76efd1645b082b0132831f4f05877a8418/src/commands/common-script.sh#L61-L68
Here $(realpath "$0") fails everytime since system-setup is not a sub-directory of linuxtoolbox

Clutchnp:
# Pull Request

## Title
Introduce Automatic Global Theming and some fixes 

## Type of Change
- [x] New feature
- [x] Bug fix
- [ ] Documentation Update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Made a global themes directory in which theme files can be stored the 3-global-theme.sh is now only used for stuff which will be common with all themes, the files change qt6ct and Kvantum themes. Also removed the creation of linuxtoolbox dir because it was causing issues 

## Testing
Tested in a latest up-to-date arch vm, running kde 

## Impact
Does what is supposed to do but may break Kde theming,we may want to check for qt6-kde and kde appropriate theming for this, this will work for most applications using qt but not gtk we may want to also look for projects  to make way here for gtk-theming

## Issue related to PR
[What issue/discussion is related to this PR (if any)]
- Resolves #122 

## Additional Information
This (referring to catppuccin theme) also sets a standard way to how themes in the future must be added

## Checklist
- [ ] My code adheres to the coding and style guidelines of the project.

### DIFF
--- a/src/commands/system-setup/3-global-theme.sh
+++ b/src/commands/system-setup/3-global-theme.sh
@@ -2,17 +2,6 @@
 
 . ./common-script.sh
 
-# Check if the home directory and linuxtoolbox folder exist, create them if they don't
-LINUXTOOLBOXDIR="$HOME/linuxtoolbox"
-
-if [ ! -d "$LINUXTOOLBOXDIR" ]; then
-    printf "${YELLOW}Creating linuxtoolbox directory: %s${RC}\n" "$LINUXTOOLBOXDIR"
-    mkdir -p "$LINUXTOOLBOXDIR"
-    printf "${GREEN}linuxtoolbox directory created: %s${RC}\n" "$LINUXTOOLBOXDIR"
-fi
-
-cd "$LINUXTOOLBOXDIR" || exit
-
 install_theme_tools() {
     printf "${YELLOW}Installing theme tools (qt6ct and kvantum)...${RC}\n"
     case $PACKAGER in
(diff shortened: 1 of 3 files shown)


infstate:
Here's the thing, you removed a lot of the intended functionality and instead added unrelated features like setting catppuccin mocha theme. The point of this module is to setup a global theme with qt6ct, qt, kvantum with a specific generic theme. Here you are doing unnecessary things like downloading files from github and using a theme that could break other things or produce unwanted affects on the system. Also the selecting themes part is completely not needed.

@ChrisTitusTech I don't approve of this PR as of now.

Clutchnp:
@infstate i don't think I have removed any functionality (if you think so can you tell me which? and I will add them back) I have only removed the creation of the linuxtoolbox directory which would have no use as mentioned in #122 if [3-global-theme.sh](https://github.com/ChrisTitusTech/linutil/blob/main/src/commands/system-setup/3-global-theme.sh) would have run, before this we were setting the general default them on all of the systems in which  [3-global-theme.sh](https://github.com/ChrisTitusTech/linutil/blob/main/src/commands/system-setup/3-global-theme.sh) was run. What I wanted to achieve with this PR was extending this functionality to many themes which can contributed over time, For some reference I ported the catppuccin theme as that is what I use.

infstate:
@Clutchnp Hello, I see. The PR overall is good in what you are trying to achieve so I approve of this now. I would say that maybe there could be some changes on how the program does this but that can be improved further on. My reasoning was because since we are supporting a lot of systems, one process may not work on all environments and that a generic theme that works on the most systems would be a lot easier for now. However this can be improved and your idea is good.  Overall @ChrisTitusTech can decide on the final decision.

Clutchnp:
I have one more concern, since *ANY* theme set by qt6ct and kvantum will conflict with a DEs theme and evironment variables(for eg: Kde and lxqt), we have 2 options, 
1. We warn the user to not use this option with DE (meaning this will only be used by people using window managers for quickly setting up their themes)  
2. We port themes on case by case basis for major DEs

lj3954:
Why in the world would we download a script that's contained within this repository and pipe it into the shell? Source it directly with `. ./system-setup/3-global-theme.sh` instead. Or, if the variables set within it need to be contained to that script, you might be able to execute it directly with `./system-setup/3-global-theme.sh` (I'm not sure whether this include_dir & tempdir solution from @JustLinuxUser preserves permissions, I doubt he knows either).

Clutchnp:
Yes you are right, i remember trying that when writing the script, but in the end it not working out, Must be drunk or sth while writing that, anyway fixed it. Thanks!

JustLinuxUser:
> Why in the world would we download a script that's contained within this repository and pipe it into the shell? Source it directly with `. ./system-setup/3-global-theme.sh` instead. Or, if the variables set within it need to be contained to that script, you might be able to execute it directly with `./system-setup/3-global-theme.sh` (I'm not sure whether this include_dir & tempdir solution from @JustLinuxUser preserves permissions, I doubt he knows either).

First, no need to be rude. But yes I don't know weather the permissions are set or not, but it doesn't actually matter, because instead of `./file` you can just write `sh file`

lj3954:
Perhaps we could, in the future, completely exclude this category from being shown when run on DEs. The XDG_CURRENT_DESKTOP environment variable can be used, alongside a blacklist of DEs (e.g. GNOME, KDE, ...). That's just an idea for the future, though.

Clutchnp:
That's a good idea, will work on it and see how that goes

ChrisTitusTech:
This shouldn't of changed the Global themeing or removal of kvantum. It should have been its own theme in a new subfolder.

Clutchnp:
@ChrisTitusTech  this doesn't remove kvantum, also it does add them in a new folder  as you are saying it should

PR #136, opened 2024-08-06
feat: Implement tablist (column)

lj3954:
# Pull Request

## Title
Implement a tablist

## Type of Change
- [x] New feature
- [x] Refactoring
- [x] UI/UX improvement

## Description
Most functions have been moved to AppState. It now contains a focus enum which specifies which area is focused, alongside containing the command or preview if necessary. Navigation is accomplished using left/right/up/down keys, or their vi equivalents.

## Testing
Behaves as expected when run. Search works correctly (searches through all tabs, including hopefully directories). 

## Impact
A tablist will allow for many more features to be added, while simplifying navigation for the user.

## Issue related to PR
- Resolves #84

## Additional Information
The UI is no good, that'll need to be worked on over time. The tabs are just a simple list for now, much like the commands on the right pane. Ideally it'd look far more like the example in the [comment](https://github.com/ChrisTitusTech/linutil/issues/84#issuecomment-2267214662). The tooltip (supposed to be in the top left) also is not yet present. 

https://github.com/user-attachments/assets/244ae22e-ffbe-4d75-9e41-02474c0bb766

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #145, opened 2024-08-09
Cleaned the main.rs file a bit.

afonsofrancof:
My previous PR #137 changed the logic for the command output float a bit.

In this PR, I am moving it out of main and into list.rs.

The logic behind this is that the only things that should belong in the main.rs file are the root widgets of the app, which for now are the search bar and the commands list.

This PR makes list.rs spawn the float of the command output, to let main be cleaner. This change also makes it so that the key press flow is cleaner.

Key presses are received by main, which then either: send it to the search bar; send it to the commands list; or process them.

The flow then continues down the widgets.
Every parent either handles the key press or passes it down to its children.

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -7,6 +7,10 @@ use ratatui::{
 pub trait FloatContent {
     fn draw(&mut self, frame: &mut Frame, area: Rect);
     fn handle_key_event(&mut self, key: &KeyEvent) -> bool;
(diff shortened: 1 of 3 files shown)


afonsofrancof:
This will have to be redone for the new tablist PR.
Might be useless now that tablist is merged, but I will check

ChrisTitusTech:
Big structure changes with merge #136

PR #146, opened 2024-08-09
Filter out special keys (arrows, page up/down, home, insert, function, end, and delete) from terminal input

tylercritchlow:
# Pull Request

## Title
Filter out special keys (arrows, page up/down, home, insert, function, end, and delete) from terminal input.

## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description
This PR filters out special keyboard keys (arrow keys, page up/down, home, end, insert, and function keys) that generate escape codes. These escape codes can interfere with terminal-based input, such as sudo passwords and confirmations, leading to unexpected behavior. By preventing these keys from being passed to the terminal, we enhance the user experience.

## Testing
I have tested each option in the linutil menu and verified that they are all working as intended.

## Impact
It may have an impact on input prompts that require these because it filters out keys and doesn't send the input for them.

## Checklist

### DIFF
--- a/src/running_command.rs
+++ b/src/running_command.rs
@@ -108,7 +108,7 @@ impl FloatContent for RunningCommand {
             KeyCode::Enter if self.is_finished() => {
                 return true;
             }
-            // do not pass arrow keys, page up/down, home, insert, any function key, or end to the terminal
+            // do not pass arrow keys, page up/down, home, insert, delete, any function key, or end to the terminal
             KeyCode::Down => {}
             KeyCode::Up => {}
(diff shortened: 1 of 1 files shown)


lj3954:
I disagree with this change, unless we have something worthwhile to do with these inputs. Some commands may depend on arrow keys. Say, some script uses a program which sometimes calls ncurses. Sending this wide variety of inputs to null doesn't sound like a good idea.

PR #248, opened 2024-09-05
Start working on GUI for this tool. 

DasMatus:
- **Start working on the GUI for this utility**
- **Comment out sidebar (will probably remove it) and start working on layout similar to winutil**

# Pull Request

## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Description

I am working on GUI that can be used on desktop distributions (e.g. Fedora), which functions similarly to the GUI that's being used to for Winutil. I am doing this, because I just want to have this utility as a desktop application. 
## Testing

I've performed a build and run test (running cargo run) to check that everything compiles. The eframe library doesn't require any external dependencies, so you can just build it.
## Impact

There are introduced two new dependencies to the codebase: [eframe](https://docs.rs/eframe) and [egui_extras](https://docs.rs/egui_extras). The first one is the UI framework that is used to make the graphical version of this program and the second one is the crate used to make Grid. 
## Issue related to PR

- Resolves #236

## Additional Information

The utility has a match arm on the beginning that checks for the $DISPLAY environment variable. 
## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #274, opened 2024-09-06
refactor: Split linutil into TUI and Core crates

lj3954:
# Pull Request

## Title
Split linutil into TUI and Core crates

## Type of Change
- [x] Refactoring

## Description
Restructure linutil from one crate containing all logic and scripts into two crates, along with a separate tabs directory to contain all data previously housed in src/commands/. Temporary directories are also entirely handled within the core crate, since they're not used anywhere else (since #153, script paths have been absolute). 

## Testing
Linutil TUI builds and functions correctly. The core crate is correctly rebuilt upon any modification to the tabs directory.

## Impact
Future contributions should be simplified, since the project is structured more clearly. Additional UIs can be added (e.g. #248) without binary bloat or build flags that heavily complicate the codebase.

Despite the appearance of large changes, current PRs should be mostly compatible with this. The only major changes are to Cargo.toml files, most of everything else is just moved to other directories.

- Closes #272 (includes modified opt-level)

## Checklist


ChrisTitusTech (approved):
Love the changes but I need to work through some of the PRs as this will break them.

adamperkowski:
Aren't there gonna be github workflows issues here?

adamperkowski:
Yes I see.

I'd suggest renaming directories for:
- `core` to `linutil-core`
- `tui` to `linutil-tui`

and removing `edition = "2021"` from `Cargo.toml`s for the crates (replacing with `edition.workspace = true`) and adding it to the main one under `[workspace.package]`.

lj3954:
> * `core` to `linutil_core`
> 
> * `tui` to `linutil_tui`

This is in the linutil repository, it's redundant to include it in the directory name. The crate names should include linutil since they could be published on crates.io or another similar repository. The guidelines you linked do not specify anything about matching the directory & crate's names.

> and removing `edition = "2021"` from `Cargo.toml`s for the crates (replacing with `edition.workspace = true`) and adding it to the main one under `[workspace.package]`.

I thought about that for a while, but I decided against it. The TUI will be far more actively maintained, and updating it to take advantage of features in new Rust editions would be more convenient if you don't have to also update completely functional code in another crate. Similarly, dependencies, even if overlap is present, aren't all being shared between the 2 crates unless is completely necessary.

adamperkowski:
> This is in the linutil repository, it's redundant to include it in the directory name. The crate names should include linutil since they could be published on crates.io or another similar repository. The guidelines you linked do not specify anything about matching the directory & crate's names.

If linutil is going to be published on crates.io (#333), there would be two separate crates (your PR) and `linutil_tui` would depend on `linhtil_core` from crates.io. I think it's better to rename the directories to avoid confusion.

> The TUI will be far more actively maintained, and updating it to take advantage of features in new Rust editions would be more convenient if you don't have to also update completely functional code in another crate.

Agreed. This makes sense.

ChrisTitusTech (approved):
This is awesome, a way better structure for future development!

PR #328, opened 2024-09-10
use dash instead of bash for inlined application-setup commands

cartercanedy:
# Use dash instead of bash for inlined application-setup commands

## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation Update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description
The setup scripts for both [mybash](https://github.com/ChrisTitusTech/mybash) & [Chris's custom neovim setup](https://github.com/ChrisTitusTech/neovim) use ANSI escapes to color text, but bash isn't posix compliant in the way that the echo command handles escapes. This change allows for proper presentation of the commands as they emit notifications without distracting escape sequences being emitted without being interpreted.

## Testing
Very simple change, just verified that the commands run with the escape sequences being interpreted properly

## Impact
Removes a lot of superfluous escape sequences and makes the command output more readable.

## Checklist

### DIFF
--- a/src/commands/applications-setup/tab_data.toml
+++ b/src/commands/applications-setup/tab_data.toml
@@ -6,7 +6,7 @@ script = "alacritty-setup.sh"
 
 [[data]]
 name = "Bash Prompt"
-command = "bash -c \"$(curl -s https://raw.githubusercontent.com/ChrisTitusTech/mybash/main/setup.sh)\""
+command = "curl -sSL https://raw.githubusercontent.com/ChrisTitusTech/mybash/main/setup.sh | sh"
 
 [[data]]
 name = "DWM-Titus"
@@ -18,7 +18,7 @@ script = "kitty-setup.sh"
 
 [[data]]
 name = "Neovim"
-command = "bash -c \"$(curl -s https://raw.githubusercontent.com/ChrisTitusTech/neovim/main/setup.sh)\""
+command = "curl -sSL https://raw.githubusercontent.com/ChrisTitusTech/neovim/main/setup.sh | sh"
 
 [[data]]
 name = "Rofi"


nnyyxxxx:
Instead of doing that I think we should just move those scripts over to this repository and make them posix compliant if they aren't.

cartercanedy:
It's probably better that they stay where they are since they're definitely going to be changing. They're posix enough that they're designed to be run with dash anyway, hence the echo issues

nnyyxxxx:
@cartercanedy I don't see why we can't just move the install scripts to this repository like we have with dwm-titus and other dot files that Chris has made.

nnyyxxxx:
> @cartercanedy I don't see why we can't just move the install scripts to this repository like we have with dwm-titus and other dot files that Chris has made.

I could probably get working on something like this.

nnyyxxxx:
I just took a look at the setup scripts in those repositories and it seems there are quite a lot of bashisms, I'm going to make a PR in a bit that will remove those bashisms and move the scripts over to this repository.

cartercanedy:
Considering that one of them is a bash config, it should be fine...

nnyyxxxx:
> Considering that one of them is a bash config, it should be fine...

@cartercanedy Running the mybash-installer doesn't mean that you're on bash; it could mean that you're using zsh or another shell, if that's the case then having bashisms inside of a script will be bad and the script won't work properly. If you take a look at the PR I created, I linked several issues containing errors from the mybash script because the person running the script was not using the bash shell and could not install mybash through the script because it contained bashisms.

nnyyxxxx:
> Considering that one of them is a bash config, it should be fine...

Running the mybash-installer doesn't have to mean you're on bash; it could mean that you're using zsh or another shell, if that's the case then having bashisms inside of a script that installs the mybash-configuration will be bad and the script won't work properly. If a person is indeed using zsh or some other shell that adheres to posix standards and does not support bashisms they will not be able to use the script to install bash and CTT's mybash config because it contained bashisms.

cartercanedy:
The setup script has no bashisms

nnyyxxxx:
@cartercanedy Also, Chris did say he wanted to eventually move everything over to this repository; so I think it should be fine.

cartercanedy:
> > The setup script for mybash* has no bashisms
> 
> The original script in CTT's repo does have bashisms
> 
> image

This warning is pay off the reason that I made the switch. Bash doesn't render escapes by default, but this is printed normally using dash or any other shell

nnyyxxxx:
> > > The setup script for mybash* has no bashisms
> > 
> > 
> > The original script in CTT's repo does have bashisms
> > image
> 
> This warning is pay off the reason that I made the switch. Bash doesn't render escapes by default, which isn't posix compliant, but this is printed normally using dash or any other shell

I see. Could you give me opinions on #335 ?

ChrisTitusTech:

PR #358, opened 2024-09-12
Updating the alacritty script

mashrukk:
**Merge [this](https://github.com/ChrisTitusTech/dwm-titus/pull/65) first, now that alacritty file in dwm-titus has changed.**


nnyyxxxx:
I don't agree with this change; having separate files is better for readability and management.

mashrukk:
@nnyyxxxx The individual files are small, so I thought combining them into one file would be a good idea. The previous script only copies the alacritty and nord files, not the keybinds, so this approach makes sense in my opinion.

nnyyxxxx:
> @nnyyxxxx The individual files are small, so I thought combining them into one file would be a good idea. The previous script only copies the alacritty and nord files, not the keybinds, so this approach makes sense in my opinion. If needed I can even remove the keybinds section to make it even smaller since it's not being used.

Again I don't think this is a very good idea, having things separated into multiple files instead of being clumped up into 1 file is always better.

PR #362, opened 2024-09-12
Add command to revert ZSH

MANICX100:
# Pull Request

## Title
Add command to revert zsh changes

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation Update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

Removes new config directory and any ZDOTDIR environment variable set. 

## Checklist


nnyyxxxx:
This seems out of place, IMO it should be done for every script. I'm sure there is a more streamline approach to this though, Instead of showing this in the TUI maybe it should be done via keybind like pressing "R" on your keyboard would undo every change made by the selected script.

adamperkowski:
I agree with @nnyyxxxx. Your approach would require creating separate `revert` scripts for basically every other script. That's not very efficient.

We could create a `revert` function in every script and do a 'revert-changes` script that would call those functions based on user input.

kleidis:
That's the only logical way to do it

Not to mention, this isn't PR-ed properly 

It's usually a good idea to avoid main -> main PRs

nnyyxxxx:
> This seems out of place, IMO it should be done for every script. I'm sure there is a more streamline approach to this though, Instead of showing this in the TUI maybe it should be done via keybind like pressing "R" on your keyboard would undo every change made by the selected script.

@lj3954 Maybe you could get working on something like this, I think it will be a great feature for Linutil just like it is in Winutil.

kleidis:
> > This seems out of place, IMO it should be done for every script. I'm sure there is a more streamline approach to this though, Instead of showing this in the TUI maybe it should be done via keybind like pressing "R" on your keyboard would undo every change made by the selected script.
> 
> @lj3954 Maybe you could get working on something like this, I think it will be a great feature for Linutil just like it is in Winutil.

Yeah, I already know users who used scripts like the global theme one and wanted to revert it afterwards because of reasons 
(This just my opinion) but it should be a bit of a priority for linutil to have this

PR #319, opened 2024-09-09
feat(fedora): add nvidia proprietary driver setup script

harshv5094:
# Pull Request

## Type of Change
- [x] New feature

## Description
This script allows fedora user to install and remove proprietary NVIDIA drivers properly, without locking their system to infinity state on startup.

## Testing
- Tested this script on my personal laptop with Fedora 40 running.
- I've run this script multiple times to check if the driver is working properly.

## Additional Information
- After installing the NVIDIA driver, before rebooting, the user need to keep their system on for up to 5 min to build drivers properly. then you need to run this command to check if your driver is installed properly or not. 
```sh
modinfo -F version nvidia
```
- So I've already implemented a solution for this in my script. I added a sleep timer for up to 4 minutes, so that the drivers can build in that time. Then it run above command to check if the driver number is showing or not.
- I've also added instruction in my script for what to do if installation doesn't work.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no errors/warnings/merge conflicts.

Issue #260, opened 2024-09-05
uninstall? undo the changes ?


icf20:
is there any way to uninstall or undo the changed made by this tool ? or any documentation on what is doing to know what to roll back

cartercanedy:
nope, this is a prerelease tool. I'd recommend taking a look at the source scripts for the util(s) that you ran and try to revert that way

ellenfel:
@icf20 
deleting the tool directly from home dir. deleting the link to `.bashrc` file  and reinstalling `.bashrc.bak` file.  did undo most if not all of the changes. iam still testing tho.

cartercanedy:
> did undo most if not all of the changes.

This is true _*only if you run the bash prompt configuration*_ or any other tools that make modifications to `~/.bashrc`, and that's only true if the configuration was run only once. This advice will otherwise not be helpful and potentially confusing.

The most generalized advice for users trying to revert changes would be to examine the shell scripts of the tools that were run to understand and restore the original configuration, if possible.

nnyyxxxx:
Yes, reverting is possible via taking a look at the scripts contents and undoing what the script did. As of now me and @adamperkowski and @lj3954 tried implementing this but found out that adding this would break a lot of systems via uninstalling certain deps. So this feature is not intended for now.

PR #370, opened 2024-09-13
feat(fedora/nvidia-*.sh):  nvidia proprietary driver setup (continued from PR #319)

harshv5094:
This PR is the continuation of #319


tabs/system-setup/fedora/nvidia-proprietary-driver-setup.sh
```diff
@@ -0,0 +1,141 @@
...
+      $ESCALATION_TOOL mv /etc/X11/xorg.conf /etc/X11/xorg.conf.saved
+    else
+      echo "/etc/X11/xorg.conf does not exist."
+    fi
+    printf "%b\n" "${GREEN}Nvidia driver was successfully removed${RC}"
+    printf "%b\n" "${YELLOW}If you want to remove the nvidia non-free repository run sudo dnf config-manager --set-disabled rpmfusion-nonfree-nvidia-driver${RC}"
+  }
+
+  echo "1. Install Nvidia Drivers"
+  echo "2. Remove Nvidia Drivers"
```
adamperkowski, line 98: No need for this. See https://github.com/ChrisTitusTech/linutil/pull/319#discussion_r1759222099, [#362  (multiple comments)](https://github.com/ChrisTitusTech/linutil/pull/362#issuecomment-2347335471) and #260.

harshv5094: I've read the PR's mentioned by you, but In case of nvidia drivers. I think the remove option is necessary for this script. Until we found any solution in the future. Because, This function not only removes the driver but also resets the changes done to `x11` files

nnyyxxxx: > I've read the PR's mentioned by you, but In case of nvidia drivers. I think the remove option is necessary for this script. Until we found any solution in the future. Because, This function not only removes the driver but also resets the changes done to `x11` files

@harshv5094 I'd still recommend removing it from this PR and storing it locally until we get around to implementing the changes mentioned in #362

harshv5094: Okay then I'll remove it

ChrisTitusTech (approved):

PR #397, opened 2024-09-15
added auto start script for desktop environments into utils

guruswarupa:
# Pull Request

## Title
added auto start script for desktop environments into utils

## Checklist

### DIFF
--- a/tabs/utils/auto_start_manager.sh
+++ b/tabs/utils/auto_start_manager.sh
@@ -1,0 +1,113 @@
+#!/bin/sh -e
+
+# Directories for application .desktop entries
+SYSTEM_APPLICATION_DIRS="/usr/share/applications /usr/local/share/applications"
+AUTOSTART_DIR="$HOME/.config/autostart"
+
+# Ensure the autostart directory exists
+mkdir -p "$AUTOSTART_DIR"
+
+# Function to list all .desktop applications from system directories
+list_all_desktop_apps() {
+    echo "Available applications:"
+    app_list=""
+
+    for dir in $SYSTEM_APPLICATION_DIRS; do
+        if [ -d "$dir" ]; then
+            for file in "$dir"/*.desktop; do
+                if [ -e "$file" ]; then
+                    app_name=$(basename "$file" .desktop)
+                    app_list="$app_list $app_name"
+                    echo "- $app_name"
+                fi
+            done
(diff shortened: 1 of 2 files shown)


tabs/utils/auto_start_manager.sh
```diff
@@ -0,0 +1,113 @@
...
+            5) echo "Exiting..."; exit 0 ;;
+            *) echo "Invalid choice. Please try again." ;;
+        esac
+
+        echo "Press [Enter] to continue..."
+        read
+    done
+}
+
+main
```
jeevithakannan2, line 113: checkEnv should be called in all scripts as per @nnyyxxxx

jeevithakannan2:
This method does not work with all desktop environments. Add a check for the current desktop environment

tabs/utils/auto_start_manager.sh
nnyyxxxx: @jeevithakannan2 Only if any of the functions in common-script is being called.

guruswarupa: uk what, stop comparing other's pull requests and use some brain on why i didn't call those 2 functions and import common-script. i didn't use any of those functions. wherein u had to use escalation_tool. u'll see my other scripts have those called as i have used escalation fn.

guruswarupa: when a reviewer asks you to do something in your pr, whats the necessity to come and type out here to do the same when you do not cross check how my script runs.

guruswarupa:
> This method does not work with all desktop environments. Add a check for the current desktop environment

yes it works on kde plasma, gnome, xfce. doesn't work for window managers

guruswarupa:
> This method does not work with all desktop environments. Add a check for the current desktop environment

could you name the desktop environment in which it doesn't work?

jeevithakannan2:
> > This method does not work with all desktop environments. Add a check for the current desktop environment
> 
> could you name the desktop environment in which it doesn't work?

Enlightment is one of the DEs i have had issues with in the past it does not work $HOME/.config/autostart


And it would be awesome if you add a warning message for users that it will not work WM

ChrisTitusTech:
Missing Standardization with common script, checks for different configurations like is it running a a supported DE, etc. 

Can't merge this.

PR #398, opened 2024-09-15
FIle organizer script 

guruswarupa:
# Pull Request

## Title
FIle organizer script for basic organizing on a directory

## Checklist

### DIFF
--- a/tabs/utils/file-organizer.sh
+++ b/tabs/utils/file-organizer.sh
@@ -1,0 +1,77 @@
+#!/bin/sh -e
+
+# Function to check if a directory exists
+check_directory() {
+    if [ ! -d "$1" ]; then
+        echo "Directory $1 does not exist. Please enter a valid directory."
+        exit 1
+    fi
+}
+
+# Function to organize files by type
+organize_by_type() {
+    TARGET_DIR="$1"
+
+    echo "Organizing files in $TARGET_DIR by type..."
+
+    for file in "$TARGET_DIR"/*; do
+        if [ -f "$file" ]; then
+            EXT=$(echo "$file" | sed 's/.*\.//')
+            if [ -z "$EXT" ]; then
+                EXT="no_extension"
+            fi
+            mkdir -p "$TARGET_DIR/$EXT"
+            mv "$file" "$TARGET_DIR/$EXT/"
+        fi
+    done
+
+    echo "Files organized by type."
+}
+
+# Function to organize files by date
+organize_by_date() {
+    TARGET_DIR="$1"
+
+    echo "Organizing files in $TARGET_DIR by date..."
+
+    for file in "$TARGET_DIR"/*; do
+        if [ -f "$file" ]; then
+            DATE=$(date -r "$file" "+%Y-%m-%d")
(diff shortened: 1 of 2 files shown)


tabs/utils/file-organizer.sh
```diff
@@ -0,0 +1,77 @@
...
+            organize_by_date "$TARGET_DIR"
+            ;;
+        *)
+            echo "Invalid choice. Please enter 1 or 2."
+            exit 1
+            ;;
+    esac
+}
+
+# Main script execution
```
jeevithakannan2, line 76: checkEnv should called at all scripts as per @nnyyxxxx

nnyyxxxx: @jeevithakannan2 Only if any of the functions in common-script is being called.

guruswarupa: 😂

jeevithakannan2: Well we should be maintaining uniformity, As like you said. I still don't get why checkEnv should be be called with all those other functions while I need to check only for escalation, the escalation function does not contain variables or any sort of things related to other functions. That's why I am so confused

nnyyxxxx: > Well we should be maintaining uniformity, As like you said. I still don't get why checkEnv should be be called with all those other functions while I need to check only for escalation, the escalation function does not contain variables or any sort of things related to other functions. That's why I am so confused

I have explained it to you many times

> Only if any of the functions in common-script are being called. Then you should call checkEnv.

jeevithakannan2: > > Well we should be maintaining uniformity, As like you said. I still don't get why checkEnv should be be called with all those other functions while I need to check only for escalation, the escalation function does not contain variables or any sort of things related to other functions. That's why I am so confused
> 
> I have explained it to you many times
> 
> > Only if any of the functions in common-script are being called. Then you should call checkEnv.

What I am asking is the escalation function does not depend on checEnv and other function, then what is purpose of calling other functions? You didn't explain why we are calling other functions all you said was it should be called and don't debate

guruswarupa: continue the debate in discord please.

jeevithakannan2: I am not debating friend, I just want to learn why we should be calling those functions if we are not going to use them anywhere. For example calling checkAUR is not needed here and checksuper group is also not useful here either.

guruswarupa: > I am not debating friend, I just want to learn why we should be calling those functions if we are not going to use them anywhere. For example calling checkAUR is not needed here and checksuper group is also not useful here either.

i had same doubt but i had to use it.

jeevithakannan2: > > I am not debating friend, I just want to learn why we should be calling those functions if we are not going to use them anywhere. For example calling checkAUR is not needed here and checksuper group is also not useful here either.
> 
> i had same doubt but i had to use it.

See I have the same doubt that's why I was asking him

ChrisTitusTech:
A bit too obscure and there are better GUI tools for this.

PR #433, opened 2024-09-17
Incorporate colors into current scripts

nnyyxxxx:
# Merge #432 before this.
# This PR is an extension of #432
## Type of Change
- [x] New feature
- [x] Refactoring
- [x] UI/UX improvement

## Description
Scripts actually use colors now, before a lot of scripts were lacking where colors were needed; now they're not.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #382, opened 2024-09-14
Rework aurhelper functionality

nnyyxxxx:
# Merge #282 before this!! very important!!
## Type of Change
- [x] New feature
- [x] Refactoring

## Description
Check this out Chris!

I'd recommend trying it out before merging just to see if you like it or not (the prompt is ugly ik, we could probably add some sort of color to it and format it differently).

## Information
- The prompt could get reworked if this gets merged, I don't have any ideas right now so I'll make a PR if this one gets merged.
- This includes functionality for not just installing an AUR helper through aurhelper-setup.sh but also everywhere that checkAURHelper is! I know you'll love this feature Chris!
- The installation of AUR Helpers only gets called if there are no AUR Helpers on the system (Paru, Yay) && the packager is pacman.

## Testing
Tested.

## Impact
Minimizes code, minimizes files, creates a interactive prompt for the user to select which aur helper to install all in 1 script.

## Issue related to PR

- Closes #374
- Closes #378 
- Resolves #377 
- Resolves #413 
- Resolves #415 

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #367, opened 2024-09-13
Add version checking & config migration for Alacritty

adamperkowski:
# Add a version check & config migration for Alacritty<br>Merge PR [#66](https://github.com/ChrisTitusTech/dwm-titus/pull/66) on dwm-titus first.

## Type of Change
- [x] New feature
- [x] Bug fix

## Description
Alacritty supports TOML config from version 0.13.0 and above. [Changelog](https://alacritty.org/changelog_0_13_0.html#Changed)
Ref #364

Auto migrate from YAML to TOML for version newer or equal to 0.13.0 of alacritty

## Issue related to PR
- Resolves #364 

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #453, opened 2024-09-18
One Terminal Script

mashrukk:
## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,7 +1,7 @@
 ## Type of Change
 - [ ] New feature
 - [ ] Bug fix
-- [ ] Documentation Update
+- [ ] Documentation update
 - [ ] Refactoring
 - [ ] Hotfix
 - [ ] Security patch
@@ -16,7 +16,7 @@
 ## Impact
 <!--[Discuss the impact of your changes on the project. This might include effects on performance, new dependencies, or changes in behaviour.]-->
 
-## Issue related to PR
+## Issues / other PRs related
 <!--[What issue/discussion is related to this PR (if any)]-->
 - Resolves #
 


tabs/applications-setup/tab_data.toml
```diff
@@ -1,8 +1,8 @@
 name = "Applications Setup"
 
 [[data]]
-name = "Alacritty"
-script = "alacritty-setup.sh"
+name = "Terminals"
```
adamperkowski, line 4:
```suggestion
name = "Terminal Emulators"
```

tabs/applications-setup/terminal-setup.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/sh -e
+
+. ../common-script.sh
+
+chooseTerminal() {
+    echo "Choose your preferred terminal:"
```
adamperkowski, line 6:
```suggestion
    echo "Choose your preferred terminal emulator:"
```

tabs/applications-setup/terminal-setup.sh
```diff
@@ -0,0 +1,82 @@
...
+        1)
+            setupAlacritty
+            setupAlacrittyConfig
+            ;;
+        2)
+            setupKitty
+            setupKittyConfig
+            ;;
+        *)
+            printf "%b\n" "${RED}Invalid choice. Please choose one of the options available.${RC}" >&2
```
adamperkowski, line 22:
```suggestion
            printf "%b\n" "${RED}Invalid choice. Please choose one of the options available.${RC}"
```

adamperkowski (changes requested):

adamperkowski:
This would break some PRs. I'd say you should wait until #433 #367 and other ones that conflict get merged.

mashrukk:
@adamperkowski Oh I see, i'll wait

adamperkowski (approved):

nnyyxxxx:
I don't agree with this at all

nnyyxxxx:
These should be kept separate so they can be maintained better.

nnyyxxxx:
Having more than 2 scripts merged into 1 is yuck imo

nnyyxxxx:
if you're going to do it this way than you would have to add literally every terminal install script into this that we add into linutil in the future, eventually this will become unusable and unreadable... It's best to keep expanding install scripts out of 1 file. Now this is not the same for my rewrite of the aur helper functionality #382 This PR only includes 2 aur_helpers and is very minimal due to that its supposed to be deployed anywhere in any script for Arch Linux users as a way to use their AUR Helpers when necessary.

mashrukk:
@nnyyxxxx right now, only two terminals are being used. Alacrity and Kitty setup covers 90% of the population. I don't think Chris will add konsole or gnome terminal setup here. I understand your point, to each its own. Please check the discord message

nnyyxxxx:
@mashrukk I still don't agree that the terminal scripts should be combined into one script; that is bad and as new terminals get added it will make this PR not good and those terminals will likely have to get added in as well, this will only make the code become unmaintainable over time and also keeping things in separate scripts is better for readability and management..

PR #491, opened 2024-09-19
added ssh command automation

guruswarupa:
# Pull Request

## added ssh command automation
- few ssh commands automation such as key generation, share key to remote host, disabling password auth, run a command on a remote host, and many more useful commands.

## Checklist


nnyyxxxx (changes requested):
I think the name of the script should be changed to something like: "Configure SSH" or something similar, since we already have a samba SSH setup inside of Linutil merging this and having both of these will be confusing for the user.

guruswarupa:
> I think the name of the script should be changed to something like: "Configure SSH" or something similar, since we already have a samba SSH setup inside of Linutil merging this and having both of these will be confusing for the user.

one is in system setup and other is in util. this script is to make using ssh simpler. its not only configuring ssh.

guruswarupa:
> > I think the name of the script should be changed to something like: "Configure SSH" or something similar, since we already have a samba SSH setup inside of Linutil merging this and having both of these will be confusing for the user.
> 
> one is in system setup and other is in util. this script is to make using ssh simpler. its not only configuring ssh. the one in system setup installs ssh and samba.

you can use the ssh remote from the script too. not just configuration.

ChrisTitusTech (approved):

PR #522, opened 2024-09-19
Differential shellcheck

lincolnthalles:
## Description

This will run [differential shellcheck](https://github.com/redhat-plumbers-in-action/differential-shellcheck) whenever someone opens a pull request of a *.sh file. 

A [shellcheck extension](https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck) recommendation is also added to vscode-based editors.

Extra [directives](https://github.com/koalaman/shellcheck/wiki/Directive) can be configured in the `.shellcheckrc` file. It's possible to enforce POSIX `sh` to the whole project by adding `shell=sh` to this file. By default, shellcheck determines the shell based on the current file shebang.

<details>

<summary>Initial PR</summary>

The original PR was able to trigger GitHub's `github-advanced-security` bot, which adds some neat review comments. This feature needs the GitHub token and permission to write security events. It was removed by request.

I would like to note that this exact action is used with the token and security events permissions by the following notorious projects: 

- [systemd](https://github.com/systemd/systemd/blob/main/.github/workflows/differential-shellcheck.yml#L31) 

- [flatpak](https://github.com/flatpak/flatpak/blob/main/.github/workflows/differential-shellcheck.yml#L29) 

- [distrobox](https://github.com/89luca89/distrobox/blob/main/.github/workflows/main.yml#L93) 

- [rbenv](https://github.com/rbenv/rbenv/blob/master/.github/workflows/lint.yml#L27) 

## Sample

image

</details>


nnyyxxxx:
The bashisms workflow should stay as it gives more information on the bashism.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+    paths:
+      - "**/*.sh"
+  workflow_dispatch:
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+
+    permissions:
+      security-events: write
```
adamperkowski, line 16: This is quite invasive. Especially if you're using `redhat-plumbers-in-action/differential-shellcheck@v5` later on which is not in any way controlled by us.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
adamperkowski, line 22: Redundant.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - id: ShellCheck
+        name: Differential ShellCheck
+        uses: redhat-plumbers-in-action/differential-shellcheck@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
```
adamperkowski, line 28: This action SHOULD NOT use `secrets.GITHUB_TOKEN`. It does not need write permissions.

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+
+      - if: ${{ always() }}
```
adamperkowski, line 30: ?

.github/workflows/shellcheck.yml
```diff
@@ -0,0 +1,35 @@
...
+        name: Upload artifact with ShellCheck defects in SARIF format
+        uses: actions/upload-artifact@v4
+        with:
+          name: Differential ShellCheck SARIF
+          path: ${{ steps.ShellCheck.outputs.sarif }}
```
adamperkowski, line 35: Why use sarif? Why even upload it?

adamperkowski (changes requested):

adamperkowski:
Can't accept this. Too invasive in my opinion.

.github/workflows/shellcheck.yml
lincolnthalles: The SARIF upload is what triggers the GitHub code review bot. Most of your concerns are addressed at the action [description](https://github.com/marketplace/actions/differential-shellcheck).

.github/workflows/shellcheck.yml
lincolnthalles: It's not. This fetches all the history. The default is 1, which fetches only a single commit. See https://github.com/actions/checkout

.github/workflows/shellcheck.yml
lincolnthalles: I see where this is coming from, but this is used similarly by the following bigger projects: 

- [systemd](https://github.com/systemd/systemd/blob/main/.github/workflows/differential-shellcheck.yml#L31) 

- [flatpak](https://github.com/flatpak/flatpak/blob/main/.github/workflows/differential-shellcheck.yml#L29) 

- [distrobox](https://github.com/89luca89/distrobox/blob/main/.github/workflows/main.yml#L93) 

- [rbenv](https://github.com/rbenv/rbenv/blob/master/.github/workflows/lint.yml#L27)

ChrisTitusTech (changes requested):
Remove write privs. This is a check and they are not needed.

ChrisTitusTech:
Ended up just using a global shellcheck. Updated all code to be inline with shellcheck recommendations and I want the test to fail if ANY fails happen during the checks.

PR #555, opened 2024-09-20
Test bashisms and shell check workflow

nnyyxxxx:
nothing to see here


nnyyxxxx:
tested and found the issue, closing.

PR #568, opened 2024-09-20
feat: implement syntax highlighting for preview text

cartercanedy:
image

Tree-sitter is a permissively-licensed, open-source parser framework, well known for its integrations in many popular editors (i use nvim, btw). Luckily, they provide Rust bindings that are reasonably straightforward to use. This PR implements syntax highlighting for the source preview, making it much easier to read for people wanting to understand the code behind commands that they are running.

## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description
- Implement tree-sitter-based syntax highlighting

## Testing
Manual testing by perusing the previews of various commands

## Impact
This should make it much easier to understand the source while being presented in the preview

## Checklist

### DIFF
--- a/tui/src/floating_text.rs
+++ b/tui/src/floating_text.rs
@@ -1,5 +1,3 @@
-#![allow(soft_unstable)]
-
 use std::{
     borrow::Cow,
     collections::VecDeque,
--- a/tui/src/floating_text.rs
+++ b/tui/src/floating_text.rs
@@ -1,56 +1,153 @@
+use std::io::{Cursor, Read as _, Seek, SeekFrom, Write as _};
+
 use crate::{
     float::FloatContent,
     hint::{Shortcut, ShortcutList},
 };
-use crossterm::event::{KeyCode, KeyEvent};
+
 use linutil_core::Command;
+
+use crossterm::event::{KeyCode, KeyEvent};
+
 use ratatui::{
     layout::Rect,
     style::{Style, Stylize},
-    text::Line,
     widgets::{Block, Borders, Clear, List},
     Frame,
 };
+
+use ansi_to_tui::IntoText;
+
+use tree_sitter_bash as hl_bash;
+use tree_sitter_highlight::{self as hl, HighlightEvent};
+use zips::zip_result;
+
 pub enum FloatingTextMode {
     Preview,
     Description,
 }
+
 pub struct FloatingText {
-    text: Vec<String>,
-    mode: FloatingTextMode,
+    src: String,
     scroll: usize,
+    n_lines: usize,
+    mode: FloatingTextMode,
+}
+
+macro_rules! style {
+    ($r:literal, $g:literal, $b:literal) => {{
+        use anstyle::{Color, RgbColor, Style};
+        Style::new().fg_color(Some(Color::Rgb(RgbColor($r, $g, $b))))
+    }};
+}
+
+const SYNTAX_HIGHLIGHT_STYLES: [(&'static str, anstyle::Style); 8] = [
+    ("function", style!(220, 220, 170)), // yellow
+    ("string", style!(206, 145, 120)),   // brown
+    ("property", style!(156, 220, 254)), // light blue
+    ("comment", style!(92, 131, 75)),    // green
+    ("embedded", style!(206, 145, 120)), // blue (string expansions)
+    ("constant", style!(79, 193, 255)),  // dark blue
+    ("keyword", style!(197, 134, 192)),  // magenta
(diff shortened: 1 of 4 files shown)


tui/src/floating_text.rs
```diff
@@ -1,56 +1,159 @@
...
+}
+
+macro_rules! style {
+    ($r:literal, $g:literal, $b:literal) => ({
+        use anstyle::{Color, Style, RgbColor};
+        Style::new().fg_color(Some(Color::Rgb(RgbColor($r, $g, $b))))
+    })
+}
+
+const SYNTAX_HIGHLIGHT_STYLES: [(&'static str, anstyle::Style); 8] = [
```
adamperkowski, line 44:
```suggestion
const SYNTAX_HIGHLIGHT_STYLES: [(&str, anstyle::Style); 8] = [
```

tui/src/floating_text.rs
```diff
@@ -1,56 +1,159 @@
...
+    ("string",      style!(206, 145, 120)), // brown
+    ("property",    style!(156, 220, 254)), // light blue
+    ("comment",     style!(92,  131, 75)),  // green
+    ("embedded",    style!(206, 145, 120)), // blue (string expansions)
+    ("constant",    style!(79,  193, 255)), // dark blue
+    ("keyword",     style!(197, 134, 192)), // magenta
+    ("number",      style!(181, 206, 168))  // light green
+];
+
+fn get_highlighted_string<'a>(s: &'a str) -> Option<String> {
```
adamperkowski, line 55:
```suggestion
fn get_highlighted_string(s: &str) -> Option<String> {
```

adamperkowski (changes requested):
Awesome.
Just remember to run `cargo fmt --all`.

cartercanedy:
> Just remember to run `cargo fmt --all`.

:))

adamperkowski (approved):

adamperkowski:
> and remove Cargo.lock

Don't remove it if you're adding deps.
If it doesn't conflict, then it's fine.

cartercanedy:
> > and remove Cargo.lock
> 
> Don't remove it if you're adding deps.
> If it doesn't conflict, then it's fine.

Did the Cargo.lock ci check get removed?

adamperkowski:
> Did the Cargo.lock ci check get removed?

It did because `Cargo.lock` exists for some reason.
Just don't push a deleated and re-cleanbuilded `Cargo.lock` if not neccessary.

cartercanedy:
>Just don't push a deleated and re-cleanbuilded `Cargo.lock` if not neccessary

makes sense

Real-MullaC:
Please take a look at the check.

cartercanedy:
> Please take a look at the check.

already on it

lj3954:
This causes a reversion by removing functionality added in #202.

cartercanedy:
Yeah, that'll have to get sorted out. I'll have to take a deeper look into breaking up the lines on overflow, since it gets a little complicated with the unrendered ansi sequences in the raw strings

cartercanedy:
Just perusing the docs a bit, this shouldn't be too hard. I'd probably still use a flat map, just over the constructed lines rather than the raw text to preserve the styling and accurately determine the width of the actually rendered text

cartercanedy:
@lj3954 why are we looking at wrapping rather than side scrolling?

ChrisTitusTech:
> @lj3954 why are we looking at wrapping rather than side scrolling?

Oh I like side scrolling a lot better.

cartercanedy:
> Oh I like side scrolling a lot better. 

Me 2, but there's also the issue of altered semantics of a mid line break without also inserting a talking backslash, and even then I'm not sure that backslashes are a perfect solution

cartercanedy:
@ChrisTitusTech @lj3954 just implemented side scrolling to fix regression

adamperkowski (approved):
`v_scroll` and `h_scroll` could be renamed to `scroll_v` and `scroll_h` but it's not that big.

lj3954:
get_lines should not take in an owned String, the split method is implemented for str instead. The lines() method also is more idiomatic than split("\n").

cartercanedy:
> get_lines should not take in an owned String, the split method is implemented for str instead. The lines() method also is more idiomatic than split("\n").

good catch, that was pretty sloppy

ChrisTitusTech (approved):

PR #569, opened 2024-09-20
Fixed some Rust code structure and modified the workflow to fail if clippy returns warnings

adamperkowski:
## Type of Change
- [x] Refactoring

## Description
Changed the unnecessary `then` & `unwrap_or` to an if statement.
Modified the Rust workflow to fail if clippy returns warnings

## Checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -38,7 +38,7 @@ jobs:
           restore-keys: ${{ runner.os }}-cargo-index-
 
       - name: Run cargo clippy
-        run: cargo clippy
+        run: cargo clippy -- -Dwarnings
 
       - name: Run cargo fmt
         run: cargo fmt --all --check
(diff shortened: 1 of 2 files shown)


lj3954:
Making the title variable mutable and appending the " [Multi-Select]" str afterwards  would probably be more readable.

adamperkowski:
> Making the title variable mutable and appending the " [Multi-Select]" str afterwards would probably be more readable.

Doing it like this may be less readable but it's better for the code. `title` doesn't need to be mutable.

lj3954:
If you want to avoid mutability, you could perhaps try something like this instead: https://github.com/lj3954/linutil/commit/8d7f2b7455b9183ab70c40d9a58a58591d4f4938.

adamperkowski:
> If you want to avoid mutability, you could perhaps try something like this instead: [lj3954@8d7f2b7](https://github.com/lj3954/linutil/commit/8d7f2b7455b9183ab70c40d9a58a58591d4f4938).

Good idea. Thanks.

lj3954 (approved):

ChrisTitusTech (approved):

PR #612, opened 2024-09-22
Enable RPMFusion repos as well on rpmfusion script

kleidis:
## Type of Change
- [x] Bug fix 

## Description
Is there any reason for the rpm fusion setup not enabling the repos?

## Testing
I was working on the Nvidia-driver script (on a non-standard fedora install)  and noticed that it was failing cuz rpm-fussion never got enabled globally and the rpm-fusion-nvidia repo only exists on fedora DE spins

## Checklist
- [ ] My code adheres to the coding and style guidelines of the project.
- [ ] My changes generate no errors/warnings/merge conflicts.

### DIFF
--- a/core/tabs/system-setup/fedora/rpm-fusion-setup.sh
+++ b/core/tabs/system-setup/fedora/rpm-fusion-setup.sh
@@ -11,7 +11,9 @@ installRPMFusion() {
                 printf "%b\n" "${YELLOW}Installing RPM Fusion...${RC}"
                 "$ESCALATION_TOOL" "$PACKAGER" install https://mirrors.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://mirrors.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
                 "$ESCALATION_TOOL" "$PACKAGER" config-manager --enable fedora-cisco-openh264
-                printf "%b\n" "${GREEN}RPM Fusion installed${RC}"
+                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-nonfree-updates
+                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-free-updates
+                printf "%b\n" "${GREEN}RPM Fusion installed and enabled${RC}"
             else
                 printf "%b\n" "${GREEN}RPM Fusion already installed${RC}"
             fi
--- a/core/tabs/system-setup/fedora/rpm-fusion-setup.sh
+++ b/core/tabs/system-setup/fedora/rpm-fusion-setup.sh
@@ -11,8 +11,8 @@ installRPMFusion() {
                 printf "%b\n" "${YELLOW}Installing RPM Fusion...${RC}"
                 "$ESCALATION_TOOL" "$PACKAGER" install https://mirrors.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://mirrors.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
                 "$ESCALATION_TOOL" "$PACKAGER" config-manager --enable fedora-cisco-openh264
-                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-nonfree-updates
-                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-free-updates
+                "$ESCALATION_TOOL" "$PACKAGER" config-manager --set-enabled rpmfusion-nonfree-updates
+                "$ESCALATION_TOOL" "$PACKAGER" config-manager --set-enabled rpmfusion-free-updates
                 printf "%b\n" "${GREEN}RPM Fusion installed and enabled${RC}"
             else
                 printf "%b\n" "${GREEN}RPM Fusion already installed${RC}"


core/tabs/system-setup/fedora/rpm-fusion-setup.sh
```diff
@@ -9,15 +9,28 @@ installRPMFusion() {
...
+                    "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-free-updates
+                    printf "%b\n" "${GREEN}RPM Fusion enabled${RC}"
+                    ;;
+                n | N)
+                    printf "%b\n" "${RED}Skipping RPM Fusion enablement${RC}"
+                    ;;
+                *)
+                    printf "%b\n" "${RED}Invalid Option! RPM Fusion not enabled${RC}"
+                    ;;
+            esac
```
adamperkowski, line 33: Doesn't enable for invalid option anyway.

```suggestion
            printf "%b\n" "${YELLOW}Do you want to enable RPM Fusion? (y/N): ${RC}"
            read -r choice
            case "$choice" in
                y|Y)
                    "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-nonfree-updates
                    "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-free-updates
                    printf "%b\n" "${GREEN}RPM Fusion enabled${RC}"
                    ;;
                *)
                    printf "%b\n" "${RED}Skipping RPM Fusion enablement${RC}"
                    ;;
            esac
```

nnyyxxxx (changes requested):
Chris doesn't like interactive scripts.

kleidis:
> Chris doesn't like interactive scripts. If you're going to be making the RPM fusion script interactive then why break the synergy with every other script; you would have to refactor this PR to make every script interactive instead of just 1.

The flathub script still has the prompt btw i might remove it

nnyyxxxx:
> > Chris doesn't like interactive scripts. If you're going to be making the RPM fusion script interactive then why break the synergy with every other script; you would have to refactor this PR to make every script interactive instead of just 1.
> 
> The flathub script still has the prompt btw i might remove it

scripts with multiple things being installed at once **need** an interactive prompt while others do not, the RPM fusion script is a prime example here, it does 1 thing, why make it interactive if its only doing 1 thing?

jeevithakannan2:
I don't think this is a necessary change !!

adamperkowski:
> Chris doesn't like interactive scripts.

@kleidis, #593 might resolve this from a user information standpoint,.

adamperkowski (approved):

kleidis:
> I don't think this is a necessary change !!

You mean 

> > Chris doesn't like interactive scripts.
> 
> @kleidis, #593 might resolve this from a user information standpoint.

That's a really nice change

nnyyxxxx (approved):

core/tabs/system-setup/fedora/rpm-fusion-setup.sh
```diff
@@ -11,7 +11,9 @@ installRPMFusion() {
                 printf "%b\n" "${YELLOW}Installing RPM Fusion...${RC}"
                 "$ESCALATION_TOOL" "$PACKAGER" install https://mirrors.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://mirrors.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
                 "$ESCALATION_TOOL" "$PACKAGER" config-manager --enable fedora-cisco-openh264
-                printf "%b\n" "${GREEN}RPM Fusion installed${RC}"
+                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-nonfree-updates
+                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-free-updates
```
nnyyxxxx, line 15: USE PACKAGER

kleidis: I'm apparently blind, nice

nnyyxxxx, line 15:
```suggestion
                "$ESCALATION_TOOL" "$PACKAGER" config-manager --set-enabled rpmfusion-free-updates
```

core/tabs/system-setup/fedora/rpm-fusion-setup.sh
```diff
@@ -11,7 +11,9 @@ installRPMFusion() {
                 printf "%b\n" "${YELLOW}Installing RPM Fusion...${RC}"
                 "$ESCALATION_TOOL" "$PACKAGER" install https://mirrors.rpmfusion.org/free/fedora/rpmfusion-free-release-$(rpm -E %fedora).noarch.rpm https://mirrors.rpmfusion.org/nonfree/fedora/rpmfusion-nonfree-release-$(rpm -E %fedora).noarch.rpm
                 "$ESCALATION_TOOL" "$PACKAGER" config-manager --enable fedora-cisco-openh264
-                printf "%b\n" "${GREEN}RPM Fusion installed${RC}"
+                "$ESCALATION_TOOL" dnf config-manager --set-enabled rpmfusion-nonfree-updates
```
nnyyxxxx, line 14:
```suggestion
                "$ESCALATION_TOOL" "$PACKAGER" config-manager --set-enabled rpmfusion-nonfree-updates
```

nnyyxxxx (changes requested):

adamperkowski:
> That's a really nice change

Tahank you.

kleidis:
Since we are here. My original pr idea was to add the option for NVIDIA open kernel modules to the driver script

Is that a good idea

nnyyxxxx:
> Since we are here. My original pr idea was to add the option for NVIDIA open kernel modules to the driver script
> 
> Is that a good idea

make it a new pr

jeevithakannan2:
Create a separate PR by modifying the nvidia-fedora script. You can make that interactive to prompt the user whether to install open kernel modules and proprietary one.

kleidis:
> > > Chris doesn't like interactive scripts. If you're going to be making the RPM fusion script interactive then why break the synergy with every other script; you would have to refactor this PR to make every script interactive instead of just 1.
> > 
> > 
> > The flathub script still has the prompt btw i might remove it
> 
> scripts with multiple things being installed at once **need** an interactive prompt while others do not, the RPM fusion script is a prime example here, it does 1 thing, why make it interactive if its only doing 1 thing?


> > Since we are here. My original pr idea was to add the option for NVIDIA open kernel modules to the driver script
> > Is that a good idea
> 
> make it a new pr

Ik, just wanted confirmation if it's a good idea to try because the project changes so fast

nnyyxxxx:
@kleidis just make it a separate script instead of combining allat into 1

adamperkowski:
> Ik, just wanted confirmation if it's a good idea to try because the project changes so fast

Yep, good to go.

kleidis:
> @kleidis just make it a separate script instead of combining allat into 1

But since you can select multiple scripts. What happens if both open and proprietary scripts are selected?

Do we have any way to disable multi-select for a certain script?

nnyyxxxx:
> > @kleidis just make it a separate script instead of combining allat into 1
> 
> But since you can select multiple scripts. What happens if both open and proprietary scripts are selected?
> 
> Do we have any way to disable multi-select for a certain script?

why would someone select both open and prop drivers to install?

kleidis:
> > But since you can select multiple scripts. What happens if both open and proprietary scripts are selected?
> > Do we have any way to disable multi-select for a certain script?
> 
> why would someone select both open and prop drivers to install?

Never underestimate what users can break, xd

I assume this also works for nvidia-open anyway so we should be goof

```
  if modinfo -F version nvidia >/dev/null 2>&1; then
```

adamperkowski:
> But since you can select multiple scripts. What happens if both open and proprietary scripts are selected?

It tries to install and conflicts.

> Do we have any way to disable multi-select for a certain script?

We have `data.preconditions` but it doesn't solve the issue of multi-selecting conflicting scripts. I think merging this into one might be a better idea after all. @nnyyxxxx?

kleidis:
> > But since you can select multiple scripts. What happens if both open and proprietary scripts are selected?
> 
> It tries to install and conflicts.
> 
> > Do we have any way to disable multi-select for a certain script?
> 
> We have `data.preconditions` but it doesn't solve the issue of multi-selecting conflicting scripts. I think merging this into one might be a better idea after all. @nnyyxxxx?

yeh i'm aware of `data.preconditions` but as you said won't work

I think this should suffice for nvidia-open 
```
  if modinfo -F version nvidia >/dev/null 2>&1; then
```

jeevithakannan2:
We don't want two separate scripts for nvidia open and proprietary. Do both of them in one script. Prompt the user of open or proprietary

kleidis:
> We don't want two separate scripts for nvidia open and proprietary. Do both of them in one script. Prompt the user of open or proprietary

Since it's 2-1 ill go with that approach than
Thanks for feedback all

adamperkowski:
No problem.

ChrisTitusTech (approved):

PR #630, opened 2024-09-22
fix global theming

### LINKED ISSUE #165: Global Theme doesn't apply everywhere
## Describe the bug
after applying "Global Dark Theme" KDE Settings is in light mode

## To Reproduce
open KDE Settings

## Expected behavior
KDE Settings to be in dark mode

## Screenshots
image

nnyyxxxx:
## Type of Change
- [x] Bug fix

## Description
- fixes KDE' theming being bricked
- Adds actual theming for specific DE's

## Testing
tested
image

## Issues / other PRs related

- Resolves #165 

## Checklist

### DIFF
--- a/core/tabs/system-setup/global-theme.sh
+++ b/core/tabs/system-setup/global-theme.sh
@@ -24,6 +24,26 @@ install_theme_tools() {
     esac
 }
 
+applyTheming() {
+    printf "%b\n" "${YELLOW}Applying global theming...${RC}"
+    case "$XDG_CURRENT_DESKTOP" in
+        KDE)
+            lookandfeeltool -a org.kde.breezedark.desktop
+            successOutput
+            exit 0
+            ;;
+        GNOME)
+            gsettings set org.gnome.desktop.interface gtk-theme "Adwaita-dark"
+            gsettings set org.gnome.desktop.interface icon-theme "Adwaita"
+            successOutput
+            exit 0
+            ;;
+        *)
+            return
+            ;;
+    esac
+}
+
 configure_qt6ct() {
     printf "%b\n" "${YELLOW}Configuring qt6ct...${RC}"
     mkdir -p "$HOME/.config/qt6ct"
@@ -46,7 +66,7 @@ EOF
 }
 
 configure_kvantum() {
-    printf "%b\n" "${YELLOW}Configuring Kvantum...${RC}\n"
+    printf "%b\n" "${YELLOW}Configuring Kvantum...${RC}"
     mkdir -p "$HOME/.config/Kvantum"
     cat <<EOF > "$HOME/.config/Kvantum/kvantum.kvconfig"
 [General]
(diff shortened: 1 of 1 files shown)


adamperkowski (approved):

GreatNovaDragon:
This is a terrible fix cause 

a) it doesnt actually fix what causes this, and 
b) instead reverts everything that the script does by overwriting it with the KDE settings. 

Kvantum may get configured, but then straight up gets ignored cause the themeing is set away from Kvantum to breeze-dark

GreatNovaDragon:
this still leaves non-KDE theme-ing broken with that.

nnyyxxxx:
the fix is fine, i'm going to re-write it for my sanity though

GreatNovaDragon:
Wouldn't be fixed yet.

The big intention of the script is "Applications should use a consistent global theme"

The script before the pr says "QT apps should use the theme the qt6 controll center tells it to use" and then sets qt6ct to say "use whatever kvantum says" and then tells kvantum "Use Breeze"
There is no Breeze kvantum theme. There also will never be one.

On QT Apps, this is also multiple steps to many, cause you can tell it directly "use breeze" instead. Would fix an issue, even though generally in KDE/QT Theming loves to use kvantum. 

It also completly ignores that GTK exists. For GTK, breeze-gtk exists in an official capability, so that could just be used there too. Yay for consistent themeing, which this script wants to achieve. 


The script after the pr says 

"Yeah, set up all the faulty stuff, but if we use KDE, throw out everything and use breeze-dark, and if we use Gnome, still use the faulty stuff in QT applications, and GTK may use adwaita-dark"

nnyyxxxx:
> Wouldn't be fixed yet.
> 
> The big intention of the script is "Applications should use a consistent global theme"
> 
> The script before the pr says "QT apps should use the theme the qt6 controll center tells it to use" and then sets qt6ct to say "use whatever kvantum says" and then tells kvantum "Use Breeze" There is no Breeze kvantum theme. There also will never be one.
> 
> On QT Apps, this is also multiple steps to many, cause you can tell it directly "use breeze" instead. Would fix an issue, even though generally in KDE/QT Theming loves to use kvantum.
> 
> It also completly ignores that GTK exists. For GTK, breeze-gtk exists in an official capability, so that could just be used there too. Yay for consistent themeing, which this script wants to achieve.
> 
> The script after the pr says
> 
> "Yeah, set up all the faulty stuff, but if we use KDE, throw out everything and use breeze-dark and ignore gtk, and if we use Gnome, still use the faulty stuff in QT applications, and GTK may use adwaita-dark"

wait till you find out you cant directly set kde's dark theme without ``lookandfeeltool``

GreatNovaDragon:
it doesnt fix global themeing though. It fixes themeing in KDE and adds an half-measure for gnome.

nnyyxxxx:
> it doesnt fix global themeing though. It fixes themeing in KDE and adds an half-measure for gnome.

again, i'm re-writing the fix to quell your concerns. Gnome is fine how it is, as it uses gtk for mostly everything.

The issue with kvantum and qt6ct will be fixed in the next commit :ok_hand:

GreatNovaDragon:
The commit doesnt fix that, it just reorders stuff. 

The solution would probably be to not install kvantum at all, and tell qt6ct to use breeze-dark where it is told to use kvantum.

nnyyxxxx:
> The commit doesnt fix that, it just reorders stuff.
> 
> The solution would probably be to not install kvantum at all, and tell qt6ct to use breeze-dark where it is told to use kvantum.

the commit fixes everything. not sure where your concerns are?

GreatNovaDragon:
my concern is that Kvantum does NOT have a breeze theme and still sets that.

nnyyxxxx:
if you're on kde or gnome the rest does not get ran, likewise if you're on any other DE the rest does get ran. More DE's can be added in the future.

GreatNovaDragon:
dwm-titus is an explicit thing that gets installed by linutil, dwm-titus is dependant on that fix

GreatNovaDragon:
yeah, that works too.

ChrisTitusTech (approved):

PR #633, opened 2024-09-23
Made popup title customizable. The Script's preview title is now its command name

afonsofrancof:
The popup window now takes a title when being created.
This allows us to select any title we want instead of having predefined titles.

This made it possible for the script previews to now have a custom title each (It just uses the command name from the list)

image

# Future

I also want to add word wrapping to the command descriptions because as of right now they go out of the screen and we need to scroll to see them, which is weird for a description.


adamperkowski (approved):

cartercanedy:
Pls see #724

ChrisTitusTech (approved):
Thanks, Love this!

PR #634, opened 2024-09-23
Restore behaviour of closing the pop-up with "q"

afonsofrancof:
I am so used to pressing q to quit everything.
I remember using it a lot, is there a reason as to why it was removed?
LTM if this is not desired behaviour.


jeevithakannan2:
Yeah I was thinking the same but I think it was not added in the first place. Anyways love this change

adamperkowski (approved):

lj3954 (approved):

cartercanedy:
this was removed as a way to prevent users from accidentally closing the application by accidentally double-tapping q, but this may or may not be a skill-issue

ChrisTitusTech (approved):

PR #646, opened 2024-09-23
Move linutil to its own dir

nnyyxxxx:
## Type of Change
- [x] Refactoring

## Description
moves linutil updater and linutil installer to its own directory

## Testing
tested
image
image


adamperkowski:
If we're going to remove preconditions, we need to check if linutil is installed in the scripts.

adamperkowski (changes requested):
If we're going to remove preconditions, we need to check if linutil is installed in the scripts.

PR #651, opened 2024-09-23
rename entries to keep synergy

nnyyxxxx:
## Type of Change
- [x] Refactoring

## Description
Renames the entries to keep the synergy with the rest of the entries

## Testing
before:
image
after:
image


guruswarupa:
why would you move grub theme setup to application ? and now renaming it to grub?

nnyyxxxx:
> why would you move grub theme setup to application ? and now renaming it to grub?

grub was already in applications-setup, i never moved it

guruswarupa:
> > why would you move grub theme setup to application ? and now renaming it to grub?
> 
> grub was already in applications-setup, i never moved it

Someone moved it. Rename it back and shift it back to system setup

adamperkowski (approved):

PR #655, opened 2024-09-23
Auto power profiling in arch needs to run an extra command.

guruswarupa:
## Type of Change
- [x] Bug fix

## Description
Auto power profiling in arch needs to run an extra command. added it to ensure all systems run it before setting power profile.

## Checklist

### DIFF
--- a/core/tabs/utils/power-profile.sh
+++ b/core/tabs/utils/power-profile.sh
@@ -34,6 +34,7 @@ installAutoCpufreq() {
         cd auto-cpufreq
         printf "%b\n" "${YELLOW}Running auto-cpufreq installer...${RC}"
         "$ESCALATION_TOOL" ./auto-cpufreq-installer
+        "$ESCALATION_TOOL"  auto-cpufreq --install
 
         cd ..
     fi


lj3954:
Other power management software should be offered in linutil as well. auto-cpufreq often significantly degrades performance, has questionable efficacy on modern CPUs, and may even hurt power efficiency in cases, since it's a python program that constantly runs in the background.

Modern AMD laptops are usually advised to run [power-profiles-daemon](https://gitlab.freedesktop.org/upower/power-profiles-daemon), which is preinstalled on many distros by default, and [tlp](https://github.com/linrunner/TLP) is a more feature rich option, usually considered the most effective on Intel systems (though it probably shouldn't be in linutil; configuration of the tool is extremely hardware dependent)

ChrisTitusTech (approved):

ChrisTitusTech:
Yeah I've fallen out of love with autocpufreq. I don't use it anymore for any of my laptops, but might as well merge this since it is offered by the utility, but I might remove it altogether.

PR #666, opened 2024-09-24
📃 fix: replace every `~` AND `,,` for posix compliance

adamperkowski:
## Type of Change
- [x] Bug fix
- [x] Refactoring

## Description
Using tilde [is a bashism](https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html), using `$HOME` is not.
Same with logic expressions and `,,`s.

## Checklist

(📃 is a shell script btw)


nnyyxxxx (approved):

adamperkowski:
More to do

adamperkowski:
Ready to merge

core/tabs/system-setup/arch/server-setup.sh
```diff
@@ -267,20 +267,23 @@ userinfo () {
...
+            if echo "$lower_machine_name" | grep -q '^[a-z][a-z0-9_.-]\{0,62\}[a-z0-9]$'
             then 
                     break 
             fi 
             # if validation fails allow the user to force saving of the hostname
-            read -r -p "Hostname doesn't seem correct. Do you still want to save it? (y/n)" force 
-            if [[ "${force,,}" = "y" ]]
+            printf "%b" "Hostname doesn't seem correct. Do you still want to save it? (y/N) "
+            read -r force
+            if [[ "$force" == "y" || "$input" == "Y" ]]
```
nnyyxxxx, line 281: did u test these archtitus changes?

adamperkowski: I did but `$input` is a typo. Fixing it now.

core/tabs/system-setup/arch/server-setup.sh
```diff
@@ -267,20 +267,23 @@ userinfo () {
      # Loop through user input until the user gives a valid hostname, but allow the user to force save 
     while true
     do 
-            read -r -p "Please name your machine: " name_of_machine
-            # hostname regex (!!couldn't find spec for computer name!!)
-            if [[ "${name_of_machine,,}" =~ ^[a-z][a-z0-9_.-]{0,62}[a-z0-9]$ ]]
+            printf "%b" "Please name your machine: "
+            read -r machine_name
+            
+            lower_machine_name=$(echo "$machine_name" | tr '[:upper:]' '[:lower:]')
```
nnyyxxxx, line 273: is this intentional? uppercase in hostnames are valid

nnyyxxxx: what if the user does indeed want uppercase letters in their hostname?

adamperkowski: This was in the script before. I only made is POSIX compliant so i guess this was originally intentional, but i think it's unnecessary.

nnyyxxxx: might as well change it

adamperkowski: In DNS, all honstnames are case-insensitive anyway so uppercase ones are at some point converted to lowecase.

nnyyxxxx: @adamperkowski if you're converting a portion of archtitus into posix, might as well make a separate PR and make it all posix

adamperkowski: I'll open a PR in a bit and add some more things.

nnyyxxxx: needs testing though

nnyyxxxx: i can volunteer to test your changes in that future pr if u wish

nnyyxxxx: also you should move these archtitus changes into that future pr as well

adamperkowski: > i can volunteer to test your changes in that future pr if u wish

Sure. I'll test everything myself but you can double-check. 

> also you should move these archtitus changes into that future pr as well

I will when I open the PR

adamperkowski:
@ChrisTitusTech In case you're wondering, this is OK to merge. I'll do the rest of archtitus in a later PR.

adamperkowski:
Clearing up confusion here:
- using `=~` is one heck of a bashism. can be avoided very easily
- `~/` is pretty obvious. it's an alias to `$HOME`. a part of GNU core utils
- `"${foo,,}" =~ ^[a-z][a-z0-9_.-]{0,62}[a-z0-9]$` has like 3+ bashisms so changing it to `tr '[:upper:]' '[:lower:]'` is completely fine. tr is much easier to understand, too

this PR explicitly cleans up some bashisms. logic remains unchanged (excluding this below, it was only `y` before, I added `Y`)
https://github.com/ChrisTitusTech/linutil/blob/f688f0c9dde9cb944f2b276230cdb3328bcb9ed8/core/tabs/system-setup/arch/server-setup.sh#L281

jeevithakannan2:
`~` is ok but the `if condition` changes is not necessary as whatever you do `server-setup.sh` will be a bash script. The arch iso ships only `bash`. The `sh` is symlinked to `bash`. You need to install `dash` package in order for this to work as posix which cannot be done

adamperkowski:
>  `~` to `$HOME` is okay but the `if condition` changes is not necessary as whatever you do `server-setup.sh` will be a bash script. The arch iso ships only `bash`. The `sh` is symlinked to `bash`. You need to install `dash` package in order for this to work as posix which cannot be done. And the shebang is bash.

I am 100℅ aware. But if we want to keep it fully posix compliant, let's keep it fully posix compliant.

PR #687, opened 2024-09-27
Feat: confirmation prompts

### LINKED ISSUE #686: Confirmation prompts
Some scripts can modify the user's environment without privilege escalation. Adding a confirmation prompt make unrecoverable, unintended changes less likely.

See https://github.com/ChrisTitusTech/linutil/pull/683#issuecomment-2379736118

cartercanedy:
# Add confirmation prompt to prevent accidental command execution

## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Description
Having an accidental keypress result in a significant environment change sucks. Adding a prompt to confirm the execution of a command makes it less likely that a user will have a destructive command accidentally execute and have to decipher the command source to revert the changes (if they even are revertible).

## Testing
locally, tested aborting and confirming command selections

## Impact
should prevent accidental neovim config deletions, at least :)

## Issues / other PRs related
- Resolves #686 

## Checklist

### DIFF
--- a/tui/src/confirmation.rs
+++ b/tui/src/confirmation.rs
@@ -36,7 +36,7 @@ impl ConfirmPrompt {
     }
 
     pub fn scroll_down(&mut self) {
-        if self.scroll < self.names.len() {
+        if self.scroll < self.names.len() - 1 {
             self.scroll += 1;
         }
     }
@@ -67,7 +67,8 @@ impl FloatContent for ConfirmPrompt {
             .iter()
             .skip(self.scroll)
             .map(|p| {
-                Line::from(Span::from(Cow::<'_, str>::Borrowed(p))).style(Style::default().bold())
+                let span = Span::from(Cow::<'_, str>::Borrowed(p));
+                Line::from(span).style(Style::default().bold())
             })
             .collect::<Text>();
 
@@ -80,6 +81,8 @@ impl FloatContent for ConfirmPrompt {
         self.status = match key.code {
             Char('y') | Char('Y') => ConfirmStatus::Confirm,
             Char('n') | Char('N') | Esc => ConfirmStatus::Abort,
+            Char('j') => { self.scroll_down(); ConfirmStatus::None },
+            Char('k') => { self.scroll_up(); ConfirmStatus::None },
             _ => ConfirmStatus::None,
         };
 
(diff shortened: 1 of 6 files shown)


lj3954:
Confirmation prompt doesn't need its own variant, we can use the generic float variant with slight refactoring. https://github.com/lj3954/linutil/commit/0a17f4272397964719f73d1cba2a06eac00c63e0

cartercanedy:
I think the focus specialization is fine there, it prevents us from needing to polute the `FloatContent` interface for other implementors

lj3954:
> I think the focus specialization is fine there, it prevents us from needing to polute the `FloatContent` interface for other implementors

I don't think it's "polluting" the trait. All that needs to be changed is to return an enum rather than a boolean. It's probably more clear that way, since the variants can explicitly state what action is occuring ('None', 'Close', etc) rather than having to look at comments in the trait to explain what each boolean value corresponds to.

adamperkowski (approved):
Works as expected, no issues.

NIT:
We could add some colors or other "aesthetic" things (like the one i mentioned below for example) to the popup though.

When selecting more than 9 entries, the list looks something like this:
```
1. Alacritty         
2. Android Debloater 
3. Bash Prompt       
4. Bottles           
5. DWM-Titus         
6. Docker            
7. Fastfetch         
8. Grub Theme        
9. Kitty             
10. Linutil Installer
11. Office Suite     
12. ZSH Prompt
```
We could make it check for the number of selected entries and add an extra space after the number for every 9/10 so it'd look like this:
```
9.  Kitty             
10. Linutil Installer
```

cartercanedy:
@lj3954 sorry for jumping to conclusions, I do like what you did with `FloatEvent` and making the `handle_key_event` api more fluent, but I do think that the confirmation float and potentially other potential features could benefit from specialization of `Float` for concrete types instead of being limited by `FloatContent` trait objects

cartercanedy:
how about
```
 1. ...
 ...
10. ...
```

adamperkowski:
> how about
> 
> ```
>  1. ...
>  ...
> 10. ...
> ```

Looks alright.

tui/src/confirmation.rs
```diff
@@ -0,0 +1,127 @@
...
+        let max_count_str = format!("{}", names.len());
+        let names = names
+            .iter()
+            .zip(1..)
+            .map(|(name, n)| {
+                let count_str = format!("{n}");
+                let space_str = (0..(max_count_str.len() - count_str.len()))
+                    .into_iter()
+                    .map(|_| ' ')
+                    .collect::<String>();
```
adamperkowski, line 35: Clippy is mad again.

```suggestion
                let space_str = (0..(max_count_str.len() - count_str.len()))
                    .map(|_| ' ')
                    .collect::<String>();
```

cartercanedy: yeah, I should just add `cargo clippy` to a precommit hook at this point

adamperkowski: Add `cargo clippy -- -Dwarnings` so it'll return `0` on warnings.

nnyyxxxx (changes requested):
love it

PR #697, opened 2024-09-29
fix(nala): remove autoclean from `system-cleanup.sh`

AngelosGian:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [x] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Remove autoclean from system-cleanup.sh . Nala doesn't have autoclean command.
Please read additional information

## Testing

edited the script under core/tabs/system-setup/system-cleanup.sh
saved the changes
cargo run >  System-setup > Full System Cleanup
Everything works without the error

## Issues / other PRs related

- Resolves #

## Additional Information

Autoclean is a command of apt-get that works. Unfortunately, nala doesn't have that command.
I tried an if statement to resolve this issue but it throws an syntax error that I couldn't resolve.
I though to make another case for nala but I leave it for discussion.

## Checklist
- [ ] I have performed a self-review of my own code.

### DIFF
--- a/core/tabs/system-setup/system-cleanup.sh
+++ b/core/tabs/system-setup/system-cleanup.sh
@@ -9,7 +9,6 @@ cleanup_system() {
             "$ESCALATION_TOOL" "$PACKAGER" clean
             "$ESCALATION_TOOL" "$PACKAGER" autoremove -y 
             "$ESCALATION_TOOL" du -h /var/cache/apt
-            "$ESCALATION_TOOL" "$PACKAGER" clean
             ;;
         zypper)
             "$ESCALATION_TOOL" "$PACKAGER" clean -a
--- a/core/tabs/system-setup/system-cleanup.sh
+++ b/core/tabs/system-setup/system-cleanup.sh
@@ -7,8 +7,7 @@ cleanup_system() {
     case "$PACKAGER" in
         apt-get|nala)
             "$ESCALATION_TOOL" "$PACKAGER" clean
-            "$ESCALATION_TOOL" "$PACKAGER" autoremove -y
-            "$ESCALATION_TOOL" "apt-get" autoclean
+            "$ESCALATION_TOOL" "$PACKAGER" autoremove -y 
             "$ESCALATION_TOOL" du -h /var/cache/apt
             "$ESCALATION_TOOL" "$PACKAGER" clean
             ;;
--- a/core/tabs/system-setup/system-cleanup.sh
+++ b/core/tabs/system-setup/system-cleanup.sh
@@ -8,7 +8,6 @@ cleanup_system() {
         apt-get|nala)
             "$ESCALATION_TOOL" "$PACKAGER" clean
             "$ESCALATION_TOOL" "$PACKAGER" autoremove -y
-            "$ESCALATION_TOOL" "$PACKAGER" autoclean
             "$ESCALATION_TOOL" du -h /var/cache/apt
             "$ESCALATION_TOOL" "$PACKAGER" clean
             ;;
--- a/core/tabs/system-setup/system-cleanup.sh
+++ b/core/tabs/system-setup/system-cleanup.sh
@@ -8,6 +8,7 @@ cleanup_system() {
         apt-get|nala)
             "$ESCALATION_TOOL" "$PACKAGER" clean
             "$ESCALATION_TOOL" "$PACKAGER" autoremove -y
+            "$ESCALATION_TOOL" "apt-get" autoclean
             "$ESCALATION_TOOL" du -h /var/cache/apt
             "$ESCALATION_TOOL" "$PACKAGER" clean
             ;;


jeevithakannan2 (approved):

nnyyxxxx (approved):

lj3954:
Why don't we just disregard the packager variable there and use `apt-get autoclean`? Nala depends on apt, correct?

nnyyxxxx:
> Why don't we just disregard the packager variable there and use `apt-get autoclean`? Nala depends on apt, correct?

why do that?

nnyyxxxx:
the user will always have apt / apt-get on their sys anyways. its best just to keep the packager var

> Why don't we just disregard the packager variable there and use `apt-get autoclean`? Nala depends on apt, correct?

lj3954:
> the user will always have apt / apt-get on their deb derivative anyways. its best just to keep the packager var

Yes. My suggestion is to change that one specific line from `"$ESCALATION_TOOL" "$PACKAGER" autoclean` to `"$ESCALATION_TOOL" apt-get autoclean`, since nala is missing this functionality and apt-get will be present on debian systems, as you mentioned.

nnyyxxxx:
> > the user will always have apt / apt-get on their deb derivative anyways. its best just to keep the packager var
> 
> Yes. My suggestion is to change that one specific line from `"$ESCALATION_TOOL" "$PACKAGER" autoclean` to `"$ESCALATION_TOOL" apt-get autoclean`, since nala is missing this functionality and apt-get will be present on debian systems, as you mentioned.

no need 
image

nnyyxxxx:
nala is being removed so it will default to apt via the packager above

core/tabs/system-setup/system-cleanup.sh
```diff
@@ -8,7 +8,6 @@ cleanup_system() {
         apt-get|nala)
             "$ESCALATION_TOOL" "$PACKAGER" clean
             "$ESCALATION_TOOL" "$PACKAGER" autoremove -y
-            "$ESCALATION_TOOL" "$PACKAGER" autoclean
```
nnyyxxxx, line 11: instead of removing, change the packager variable to apt-get

AngelosGian: Thank you very much. I haven't thought of that. I test it and it works. Again thank you very much.

jeevithakannan2:
Removing autoclean is the better option here

`man apt-get`

```
clean
   clean clears out the local repository of retrieved package files. It
   removes everything but the lock file from /var/cache/apt/archives/ and
   /var/cache/apt/archives/partial/.

autoclean (and the auto-clean alias since 1.1)
   Like clean, autoclean clears out the local repository of retrieved
   package files. The difference is that it only removes package files
   that can no longer be downloaded, and are largely useless. This allows
   a cache to be maintained over a long period without it growing out of
   control. The configuration option APT::Clean-Installed will prevent
   installed packages from being erased if it is set to off.
```

Doing a `apt-get clean` will remove everything in `/var/cache/apt/archives/` so `apt-get autoclean` is useless here

core/tabs/system-setup/system-cleanup.sh
```diff
@@ -8,7 +8,7 @@ cleanup_system() {
...
+            "$ESCALATION_TOOL" "apt-get" autoclean
```
jeevithakannan2:
```suggestion
```

Removing `autoclean` is the better option here

`man apt-get`

```
clean
   clean clears out the local repository of retrieved package files. It
   removes everything but the lock file from /var/cache/apt/archives/ and
   /var/cache/apt/archives/partial/.

autoclean (and the auto-clean alias since 1.1)
   Like clean, autoclean clears out the local repository of retrieved
   package files. The difference is that it only removes package files
   that can no longer be downloaded, and are largely useless. This allows
   a cache to be maintained over a long period without it growing out of
   control. The configuration option APT::Clean-Installed will prevent
   installed packages from being erased if it is set to off.
```

Doing a `apt-get clean` will remove everything in `/var/cache/apt/archives/` so `apt-get autoclean` is useless here.

`nala clean` also does the same by removing everything in `/var/cache/apt/archives/`

Adding `apt-get autoclean` will not do anything here it is basically useless if `apt-get clean` or `nala clean` is used

AngelosGian:
Well, I will leave it as is (deleted). Chris will decide if he will accept it or not. So, no big deal. Again thank you for the feedback.

nnyyxxxx:
> Removing `autoclean` is the better option here
> 
> `man apt-get`
> 
> ```
> clean
>    clean clears out the local repository of retrieved package files. It
>    removes everything but the lock file from /var/cache/apt/archives/ and
>    /var/cache/apt/archives/partial/.
> 
> autoclean (and the auto-clean alias since 1.1)
>    Like clean, autoclean clears out the local repository of retrieved
>    package files. The difference is that it only removes package files
>    that can no longer be downloaded, and are largely useless. This allows
>    a cache to be maintained over a long period without it growing out of
>    control. The configuration option APT::Clean-Installed will prevent
>    installed packages from being erased if it is set to off.
> ```
> 
> Doing a `apt-get clean` will remove everything in `/var/cache/apt/archives/` so `apt-get autoclean` is useless here.
> 
> `apt clean` or `nala clean` also does the same by removing everything in `/var/cache/apt/archives/`

well it was lj's suggestion to add it back after all :joy:

core/tabs/system-setup/system-cleanup.sh
```diff
@@ -7,8 +7,7 @@ cleanup_system() {
     case "$PACKAGER" in
         apt-get|nala)
             "$ESCALATION_TOOL" "$PACKAGER" clean
```
jeevithakannan2, line 9: Also remove this duplicate clean


```suggestion
```

jeevithakannan2 (changes requested):

core/tabs/system-setup/system-cleanup.sh
AngelosGian: Right I forgot that.

jeevithakannan2 (approved):

nnyyxxxx (approved):

lj3954 (approved):

adamperkowski (approved):
Good to merge

PR #706, opened 2024-09-30
chore: Add spacing before and after for tui titles

jeevithakannan2:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description
- Add spacing before and after titles and tips in TUI.

Before

before

After

after

## Testing
- Tested no issues.

## Additional Information
- `Box::leak` was used for cool tips as it is static and will be displayed through out the whole program.
- `Box::leak` must be used with caution as it can lead to memory leaks. But here the cool tips is not borrowed or used in mutable state so there should be no issues.

## Checklist

### DIFF
--- a/tui/src/state.rs
+++ b/tui/src/state.rs
@@ -21,7 +21,7 @@ use ratatui::{
 
 const MIN_WIDTH: u16 = 77;
 const MIN_HEIGHT: u16 = 19;
-const TITLE: &str = concat!("Linux Toolbox - ", env!("BUILD_DATE"));
+const TITLE: &str = concat!(" Linux Toolbox - ", env!("BUILD_DATE"), " ");
 const ACTIONS_GUIDE: &str = "List of important tasks performed by commands' names:
 
 D  - disk modifications (ex. partitioning) (privileged)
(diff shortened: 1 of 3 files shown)


nnyyxxxx (approved):

lj3954:
I would just return the String rather than leaking it. Or you could return a Box<str> if you want to drop the unnecessary capacity to save 8 bytes of memory. Box::leak is mainly useful if you're passing data between threads, or doing anything else the borrow checker wouldn't particularly like. There's no reason we'd need to leak this, and it's better to avoid calling functions with the potential of causing memory issues unless they're necessary.

adamperkowski (approved):

tui/src/state.rs
```diff
@@ -357,17 +357,17 @@ impl AppState {
...
         let title = if self.multi_select {
-            &format!("{} [Multi-Select]", TITLE)
+            &format!("{}[Multi-Select] ", TITLE)
         } else {
             TITLE
         };
 
         #[cfg(feature = "tips")]
-        let bottom_title = Line::from(self.tip.bold().blue()).right_aligned();
+        let bottom_title = Line::from(self.tip.clone().bold().blue()).right_aligned();
```
lj3954, line 366:
```suggestion
        let bottom_title = Line::from(self.tip.as_str().bold().blue()).right_aligned();
```

lj3954 (changes requested):

lj3954 (approved):

PR #725, opened 2024-10-02
return user to their previous position if they exit a subdir

nnyyxxxx:
# watch the video

## Type of Change
- [x] New feature

## Description
takes the user back to their previous position when they exit a subdir

tested with a subdir inside of a subdir works as expected

tested with a regular subdir works as expected as well

example:

https://github.com/user-attachments/assets/688dcae5-35b6-4c19-9b86-f700673ce14f

before, it would just take u back to the first entry in the list


tui/src/state.rs
```diff
@@ -572,9 +574,12 @@ impl AppState {
...
     fn enter_parent_directory(&mut self) {
-        self.visit_stack.pop();
-        self.selection.select(Some(0));
-        self.update_items();
+        if self.visit_stack.len() > 1 {
+            self.visit_stack.pop();
+            let previous_position = self.position_stack.pop().unwrap_or(0);
+            self.selection.select(Some(previous_position));
+            self.update_items();
+        }
```
adamperkowski, line 582:
```suggestion
        if let Some(previous_position) = self.position_stack.pop() {
            self.visit_stack.pop();
            self.selection.select(Some(previous_position));
            self.update_items();
        }
```

cartercanedy (changes requested):
Put `visit_stack` and `position_stack` together:
```rust
// ...
    visit_stack: Vec<(Rc<ListNode>, usize)>
// ...
```

### Correction:
```rust
// ...
    visit_stack: Vec<(NodeId, usize)>,
// ...
```

tui/src/state.rs
```diff
@@ -668,7 +675,11 @@ impl AppState {
     }
 
     fn handle_enter(&mut self) {
-        if self.selected_item_is_cmd() {
+        if self.selected_item_is_up_dir() {
+            self.enter_parent_directory();
+        } else if self.selected_item_is_dir() {
+            self.go_to_selected_dir();
+        } else if self.selected_item_is_cmd() {
```
adamperkowski, line 682: We could make a selected item type here and then use `match`. Just an idea, that would need a refactor.

adamperkowski (changes requested):

nnyyxxxx:
> Put `visit_stack` and `position_stack` together:
> 
> ```rust
> // ...
>     visit_stack: Vec<(Rc<ListNode>, usize)>
> // ...
> ```

its best we keep them separated

cartercanedy:
> its best we keep them separated

Besides unnecessarily making a heap allocation, they're both participating in the same functionality; they're best together

adamperkowski (approved):

tui/src/state.rs
nnyyxxxx: solved with [ce5f037](https://github.com/ChrisTitusTech/linutil/pull/725/commits/ce5f0375b0ecec3e5beed2133f8a3713718d7987)

nnyyxxxx:
> > its best we keep them separated
> 
> Besides unnecessarily making a heap allocation, they're both participating in the same functionality; they're best together

sorry but i see no performance impact with this current implementation, if we want to combine them then we can refactor it in the future. For now this implementation is just a stepping stone.

cartercanedy:
> sorry but i see no performance impact with this current implementation, if we want to combine them then we can refactor it in the future. For now this implementation is just a stepping stone.

I figure that there's no better time than now
Just opened a PR for your fork, you're free to merge the changes if you want

cartercanedy (approved):
Awesome, thanks for merging that commit

ChrisTitusTech (approved):

PR #713, opened 2024-10-01
📃 fix(monitor setup): a bug with utility-functions; newlines & colors in the scripts

adamperkowski:
## Type of Change
- [x] Bug fix
- [x] Refactoring
- [x] Hotfix

## Before and After
image
---
image
---
image
---
image
---
image
---
image

## Additional Information
I found another bug when testing this:
image

- Shows two separate monitors as one
- Clearly doesn't work

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] My changes generate no errors/warnings/merge conflicts.

(📃 is a shell script btw)

PR #732, opened 2024-10-02
fix issues with extend displays

nnyyxxxx:
## Type of Change
- [x] Bug fix

image

fixes an issue with utility functions trying to be called from the current directory instead of the previous directory


adamperkowski:
#713 duplicate :((

nnyyxxxx:
didnt see that

PR #748, opened 2024-10-03
perf(paru-setup): use `-bin` packages for faster installation

nnyyxxxx:
## Type of Change
- [x] Refactoring

## Description
uses binary instead of building from source


adamperkowski (approved):
So that's the 5th or 7th PR about the same this. Let's hope this is the last one.

nnyyxxxx:
> So that's the 5th or 7th PR about the same thing. Let's hope this is the last one.

lets all give our thanks to the ghost
image

adamperkowski:
o7

ChrisTitusTech (approved):

Issue #742, opened 2024-10-02
Arch server NOT working


angryguy-win:
## Describe the bug
arch server install NOT working

## Screenshots
![image](https://github.com/user-attachments/assets/d0937d6c-b48c-4b9e-8530-6a69fadc8581)

adamperkowski:
Well that definitely looks like an issue. Could you provide more information?
1. What formal linutil version are you running
2. What branch
3. Logs (if any)
4. The output of `sh -x server-setup.sh` after `cd`ing into `core/tabs/system-setup/arch` if you're running linutil from source

This is clearly a bug but it's pretty hard to guess by a single screenshot.

/label 'script'

angryguy-win:
git clone from main 
I'm checking and gathering what I can will get back to this...

adamperkowski:
The output of

```bash
cd core/tabs/system-setup/arch
sh -x server-setup.sh
```

would provide all info needed.

angryguy-win:
Sorry I figured it out

nnyyxxxx:
you cannot run it from the graphical interface it has to be run through the tty

PR #751, opened 2024-10-03
Mybash dependencies

### LINKED ISSUE #750: Mybash deps check is bugged
## Describe the bug
`mybash-setup.sh` uses `command_exists "bash" "bash-completion" ...` as if it fails if one of the dependencies does not exist, but this won't work as written. Also, `bash-completion` and `fontconfig` aren't executables, so the check would always fail if it did work as intended...

## To Reproduce
1. run the bash prompt script
2. install fails if deps aren't installed, most likely to happen on fresh distro installs

## Checklist

cartercanedy:
## Type of Change
- [x] Bug fix

## Description
Corrects the dependency executable tests

## Testing
tested on a fresh install of openSUSE Leap 15.6 vm

## Impact
Should make mybash run as intended on fresh installs

## Issues / other PRs related
- Resolves #750 

## Checklist

### DIFF
--- a/core/tabs/applications-setup/mybash-setup.sh
+++ b/core/tabs/applications-setup/mybash-setup.sh
@@ -4,17 +4,31 @@
 
 gitpath="$HOME/.local/share/mybash"
 
-installDepend() {
-    if ! command_exists bash bash-completion tar bat tree unzip fontconfig git; then
-    printf "%b\n" "${YELLOW}Installing Bash...${RC}"
-    case "$PACKAGER" in
-        pacman)
-            "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm bash bash-completion tar bat tree unzip fontconfig git
-            ;;
-        *)
-            "$ESCALATION_TOOL" "$PACKAGER" install -y bash bash-completion tar bat tree unzip fontconfig git
-            ;;
-    esac
+installDeps() {
+    DEPS="bash bash-completion tar bat tree unzip fontconfig git"
+    depsNeeded=""
+
+    for dep in $DEPS; do
+        if ! command_exists "${dep}"; then
+            depsNeeded="${depsNeeded} ${dep}"
+        fi
+    done
+
+    if [ "$depsNeeded" != "" ]; then
+        printf "%b\n" "${RED}Some dependencies weren't installed\nInstalling${depsNeeded}"
+
+        case "$PACKAGER" in
+            pacman)
+                # disable globbing/word-splitting warning, need word splitting
+                # shellcheck disable=SC2086
+                "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm $depsNeeded 
+                ;;
+            *)
+                # disable globbing/word-splitting warning, need word splitting
+                # shellcheck disable=SC2086
+                "$ESCALATION_TOOL" "$PACKAGER" install -y $depsNeeded 
+                ;;
+        esac
     fi
 }
 
@@ -102,7 +116,7 @@ linkConfig() {
 
 checkEnv
 checkEscalationTool
-installDepend
+installDeps
 cloneMyBash
 installFont
 installStarshipAndFzf
--- a/core/tabs/applications-setup/mybash-setup.sh
+++ b/core/tabs/applications-setup/mybash-setup.sh
@@ -6,30 +6,31 @@ gitpath="$HOME/.local/share/mybash"
 
 installDeps() {
     DEPS="bash bash-completion tar bat tree unzip fontconfig git"
-    depsNeeded=""
 
-    for dep in $DEPS; do
+    # no way to reliably test for bash-completion for all I know
+    # fontconfig isn't an executable, so use fc-cache to test for installation
+    DEP_TESTS="bash tar bat tree unzip fc-cache git"
+
+    for dep in $DEP_TESTS; do
         if ! command_exists "${dep}"; then
-            depsNeeded="${depsNeeded} ${dep}"
+            printf "%b\n" "${RED}Some dependencies weren't installed\n${GREEN}Installing${RC} [$(echo "${DEPS}" | sed 's/ /, /g')]"
+
+            case "$PACKAGER" in
+                pacman)
+                    # disable globbing/word-splitting warning, need word splitting
+                    # shellcheck disable=SC2086
+                    "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm $DEPS
+                    ;;
+                *)
+                    # disable globbing/word-splitting warning, need word splitting
+                    # shellcheck disable=SC2086
+                    "$ESCALATION_TOOL" "$PACKAGER" install -y $DEPS
+                    ;;
+            esac
+
+            break
         fi
     done
-
-    if [ "$depsNeeded" != "" ]; then
-        printf "%b\n" "${RED}Some dependencies weren't installed\nInstalling${depsNeeded}"
-
-        case "$PACKAGER" in
-            pacman)
-                # disable globbing/word-splitting warning, need word splitting
-                # shellcheck disable=SC2086
-                "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm $depsNeeded 
-                ;;
-            *)
-                # disable globbing/word-splitting warning, need word splitting
-                # shellcheck disable=SC2086
-                "$ESCALATION_TOOL" "$PACKAGER" install -y $depsNeeded 
-                ;;
-        esac
-    fi
 }
 
 cloneMyBash() {


nnyyxxxx:
you're doing so much unnecessary things here

nnyyxxxx:
there is nothing wrong with the script, not sure what you're trying to accomplish here, every pkg listed there is valid and was in the original mybash-setup script.

image

nnyyxxxx:
```bash
        *)
            "$ESCALATION_TOOL" "$PACKAGER" install -y bash bash-completion tar bat tree unzip fontconfig git
            ;;
```
if this is the issue then expand it out to more packagers && put a pkger not supported in the *)

nnyyxxxx:
though it really shouldnt be as these are highly known packages, so really there is no issue in the first place. If you could provide any screenshots on the error you encountered that would be great

nnyyxxxx (changes requested):
revert these changes, command_exists only checks if the first package exists as a command. Also there is no issue with how it is, even if the command does not exist it will still run due to the "!", this PR is pretty redundant.

nnyyxxxx:
image
also this makes zero sense, because of the ! even if 1 is installed and none of the rest are installed then it will still begin installation this includes if all are installed as well

nnyyxxxx:
there is zero issue here, just close this nonsense

nnyyxxxx:
and in your original issue, you're wrong again because multiple pkg names that are included in command_exists will be ignored, only the 1st pkg will be used for checking. Instead refactor this PR to remove all additional pkgs from the command_exists variable

nnyyxxxx:
s


all this PR does is bloat up the deps function

core/tabs/applications-setup/mybash-setup.sh
```diff
@@ -4,18 +4,33 @@
...
+                *)
+                    # disable globbing/word-splitting warning, need word splitting
+                    # shellcheck disable=SC2086
+                    "$ESCALATION_TOOL" "$PACKAGER" install -y $DEPS
+                    ;;
+            esac
+
+            break
+        fi
+    done
```
nnyyxxxx, line 33:
```suggestion
installDepend() {
    if ! command_exists bash; then
    printf "%b\n" "${YELLOW}Installing Bash...${RC}"
    case "$PACKAGER" in
        pacman)
            "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm bash bash-completion tar bat tree unzip fontconfig git
            ;;
        *)
            "$ESCALATION_TOOL" "$PACKAGER" install -y bash bash-completion tar bat tree unzip fontconfig git
            ;;
    esac
    fi
```

core/tabs/applications-setup/mybash-setup.sh
```diff
@@ -102,7 +117,7 @@ linkConfig() {
 
 checkEnv
 checkEscalationTool
-installDepend
+installDeps
```
nnyyxxxx, line 120:
```suggestion
installDepend
```

cartercanedy:
Have you tried running the command on a fresh headless installation? It fails because there are missing dependencies, particularly fc-cache that comes with fontconfig. Please try the current version deployment on a fresh VM and tell me what happens

cartercanedy:
> Have you tried running the command on a fresh headless installation? It fails because there are missing dependencies, particularly fc-cache that comes with fontconfig. Please try the current deployment on a fresh headless VM and tell me what happens

cartercanedy:
for context, here's the issue manifested.
the command fails since it's current implementation inadequately verifies that the dependencies are installed
image

nnyyxxxx:
> for context, here's the issue manifested. the command fails since it's current implementation inadequately verifies that the dependencies are installed image

again, re read what i said, the simple solution to this is to use 1 pkg for detection. All you're doing here is bloating up the deps function.

cartercanedy:
Your suggestion as implemented would not have resolved the issue

nnyyxxxx:
> Your suggestion as implemented would not have resolved the issue

yes it would resolve the issue, you're removing all other dependencies, how would a simple installation of one dep not work?

nnyyxxxx:
> Your suggestion as implemented would not have resolved the issue

explain how it wouldnt resolve the issue?

nnyyxxxx:
there literally is no issue here, just remove every other dep from the command exists line lmao

core/tabs/applications-setup/mybash-setup.sh
nnyyxxxx, line 33:
```suggestion
installDeps() {
    if ! command_exists bash; then
    printf "%b\n" "${YELLOW}Installing Bash...${RC}"
    case "$PACKAGER" in
        pacman)
            "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm bash bash-completion tar bat tree unzip fontconfig git
            ;;
        *)
            "$ESCALATION_TOOL" "$PACKAGER" install -y bash bash-completion tar bat tree unzip fontconfig git
            ;;
    esac
    fi
```

cartercanedy:
You've refuse to recreate the environment that I've already been very clear about, and are still choosing to insist on your incorrect understanding of the issue. 

It _**is**_ an issue, this fixes it without additional regressions

nnyyxxxx:
> You've refuse to recreate the environment that I've already been very clear about, and are still choosing to insist on your incorrect understanding of the issue.
> 
> It _**is**_ an issue, this fixes it without additional regressions

again there is no issue here, what you're doing is bloating up the deps function..

nnyyxxxx:
there is no need to verify if deps are installed if they are literally getting installed beforehand

nnyyxxxx:
anyways, i cant even reproduce the issue you're having. The script works perfectly fine for me

core/tabs/applications-setup/mybash-setup.sh
cartercanedy: To fix my specific issue, you would have to test for `fc-cache`, not `bash`. Each deep needs to be checked for independently. This wouldn't fix the issue, it would pretty much guarantee that the dependency specification is never properly verified or satisfied considering how many distros provide bash out of the box

cartercanedy:
> anyways, i cant even reproduce the issue you're having. The script works perfectly fine for me

You're being ignorant at this point

core/tabs/applications-setup/mybash-setup.sh
nnyyxxxx: > To fix my specific issue, you would have to test for `fc-cache`, not `bash`. Each deep needs to be checked for independently. This wouldn't fix the issue, it would pretty much guarantee that the dependency specification is never properly verified or satisfied considering how many distros provide bash out of the box

this does fix the issue. and anyways there literally is NO issue here its clear to me that no testing was involve to encounter such issue and all you're doing is trying to bloat up the deps func

nnyyxxxx:
> > anyways, i cant even reproduce the issue you're having. The script works perfectly fine for me
> 
> You're being ignorant at this point

again there is no issue here, as i said above i cant reproduce. The fontconfig package is literally getting installed, WHY check for these deps after they are installed?? you are just bloated up the func

cartercanedy:
It's incredible how hostile you're being right now

nnyyxxxx:
> It's incredible how hostile you're being right now

no hostility is being expressed, i am asking a question as seen here

> The fontconfig package is literally getting installed, WHY check for these deps after they are installed?? you are just bloating up the func

the capped "why" is to show frustration over you not getting the point i'm trying to make

cartercanedy:
The point of checking for dependencies is to preserve efficiency. An alternative is to just blindly install the dependencies without writing about checking for deps. The script fails as it's currently written; there's multiple solutions, and I can only implement one. You've repeatedly insisted that there is no issue. If that's your point, then you are either mistaken or intentionally being combative.

nnyyxxxx:
> The point of checking for dependencies is to preserve efficiency. An alternative is to just blindly install the dependencies without writing about checking for deps. The script fails as it's currently written; there's multiple solutions, and I can only implement one. You've repeatedly insisted that there is no issue. If that's your point, then you are either mistaken or intentionally being combative.

if you are trying to implement dep checking, then why arent you doing this for every script? also as ive said multiple times, if you are installing a set of packages why would they fail? Have you ever been in the process of installing a package and it fails on you? I dont think so... Going by that, there is no issue here, if you're installing fontconfig then fc-cache should work perfectly fine. no need to check for deps all that does is create unnecessary overhead and more time spent on the users end to finish the script

nnyyxxxx:
also to address the statement above about the command_exists bash, the script is literally installing BASH and is implementing bash configurations, also bash isnt the only shell, and its certainly not one that the user has to keep on their system after installation; alternatives exist: such as Zsh, nushell, fish, and multiple others.

cartercanedy:
Then make the pr yourself. @ChrisTitusTech can decide whether or not to merge. I'm not going to continue this conversation

cartercanedy:
As a final note, it's very disheartening to see how you've reacted to this bug report. I've made these changes in good faith, but your deliberately rude are not collaborative I'm the slightest. Your first words in response to my bug report were "you're wrong", and I believe make my point very clear.

nnyyxxxx:
> As a final note, it's very disheartening to see how you've reacted to this bug report. I've made these changes in good faith, but your deliberately rude are not collaborative I'm the slightest. Your first words in response to my bug report were "you're wrong", and I believe make my point very clear.

i've explained to you many times that i cannot reproduce the issue you're having, i have even gone to the lengths of firing up a VM, and could not reproduce in there either. 

and as said above

> if you are trying to implement dep checking, then why arent you doing this for every script? also as ive said multiple times, if you are installing a set of packages why would they fail? Have you ever been in the process of installing a package and it fails on you? I dont think so... Going by that, there is no issue here, if you're installing fontconfig then fc-cache should work perfectly fine. no need to check for deps all that does is create unnecessary overhead and more time spent on the users end to finish the script

Please answer this i would like to hear your thoughts, it explains what i'm trying to convey very well

cartercanedy:
I'm not going to walk through every line of code with you. Just evaluating the code statically, it was clear to me what the issue was. I implemented the fix, the issue was resolved in my edge-case environment, and trying the script still in my dev environment still worked, indicating that the program was not correct before, and is correct now. The dependencies were not being installed if not present prior to these changes, and now they are. It is an issue, it is not a "non-existent" issue, as you stated. Read my reproduction steps, it's dead simple to reproduce.

nnyyxxxx:
> I'm not going to walk through every line of code with you. Just evaluating the code statically, it was clear to me what the issue was. I implemented the fix, the issue was resolved in my edge-case environment, and trying the script still in my dev environment still worked, indicating that the program was not correct before, and is correct now. The dependencies were not being installed if not present prior to these changes, and now they are. It is an issue, it is not a "non-existent" issue, as you stated. Read my reproduction steps, it's dead simple to reproduce.

theres no issue here

nnyyxxxx:
the script works perfectly fine for me in and out of VM. Can't reproduce this issue at all

lj3954 (approved):
image

Yes, there is an issue with the previous syntax. And yes, that would cause a problem if the first command (bash) exists (which it nearly certainly will) while any of the others don't.

nnyyxxxx:
both of these versions are basically doing the same thing

1. previous version installs the deps (and does it correctly as this is in every other script
2. new version installs the deps then checks if those deps are installed

This new version bloats the script up, not sure what your issue is here @cartercanedy and i dont know what you're trying to accomplish by adding this complexity onto the script; as i've said many times the issue you're facing is unreproducable meaning i can NOT reproduce this, in and out of a VM, tested on host (Arch Linux) and in a VM (openSUSE).

This reference
image
is wrong because of the !, infact it should never fail because the ! makes it so if any of them are missing it will proceed and install and since bash-completion and fontconfig aren't executables it will always run through. Along with this i suggested removing all of the deps from the command_exists function and replacing them with just BASH but of course you declined and that is the: quote me on this "solution to your nonexistent problem" as stated in the images above.

These changes do the following:
1. cause unnecessary overhead
2. accomplish the same thing as the previous version except for dep checking (which is not necessary)
3. potential for more unnecessary maintenance burden

please just close this non-sense, i have gave you a solution and you have declined it (for reasons unknown) this is just disgusting.

nnyyxxxx:
> image
> 
> Yes, there is an issue with the previous syntax. And yes, that would cause a problem if the first command (bash) exists (which it nearly certainly will) while any of the others don't.

if that is true then why do all of the other scripts work? many people have ran through them including myself and have encountered 0 issues. Please explain

lj3954:
@nnyyxxxx There is a syntax error in the current code. We're passing multiple arguments to a function which only can accept one. If you'd like to propose instead allowing the command_exists function to check through a list of arguments rather than just one (using a loop, much like this), you may. But this is code with a syntax error, and this PR just fixes said error. We don't need this pointless tirade about "performance" and how we "don't need these changes" again, it's ridiculous.

nnyyxxxx:
> @nnyyxxxx There is a syntax error in the current code. We're passing multiple arguments to a function which only can accept one. If you'd like to propose instead allowing the command_exists function to check through a list of arguments rather than just one (using a loop, much like this), you may. But this is code with a syntax error, and this PR just fixes said error. We don't need this pointless tirade about "performance" and how we "don't need these changes" again, it's ridiculous.
> 
> The script works for you because all of these commands exist on your system. If one didn't exist on someone else's system, we'd have an issue. command_exists does not take in more than one parameter, the function is not being called correctly and all of the commands past the first are being completely disregarded.

as i have said in my previous posts in this thread i **know** that only 1 argument can be passed to command_exists and i have **suggested** the change to simplify it down to just **bash** many times, what hes doing here is completely unnecessary

nnyyxxxx:
https://github.com/ChrisTitusTech/linutil/pull/751#discussion_r1787417459
https://github.com/ChrisTitusTech/linutil/pull/751#discussion_r1788455751

lj3954:
The other programs are also dependencies, correct? "Simplifying" the dependency check to not check for dependencies isn't a solution to anything. We can't expect 'unzip' or 'fontconfig' to be installed just because bash is, those aren't dependencies of bash.

nnyyxxxx:
image

nnyyxxxx:
> The other programs are also dependencies, correct? "Simplifying" the dependency check to not check for dependencies isn't a solution to anything. We can't expect 'unzip' or 'fontconfig' to be installed just because bash is, those aren't dependencies of bash.

except it is a solution when literally **every** other script does it, if we want a solution to the 1 dep checking issue, then we can alter the command_exists function in common-script instead of modifying every script to include this dep checking nonsense

lj3954:
If every script is calling a function with the wrong amount of parameters, then they're just all wrong. Indeed, you can suggest instead modifying the command_exists function to allow passing in multiple parameters. But suggesting that you only check for one of the numerous dependencies, or claiming that there's "no issue" because you can't reproduce it yourself, isn't helpful.

nnyyxxxx:
image

nnyyxxxx:
https://github.com/ChrisTitusTech/linutil/pull/762

is an easier way to solve this "issue" without having to add dep checks to every script

cartercanedy:
> both of these versions are basically doing the same thing
> 
> 1. previous version installs the deps (and does it correctly as this is in every other script
> 2. new version installs the deps then checks if those deps are installed
> 
> This new version bloats the script up, not sure what your issue is here @cartercanedy and i dont know what you're trying to accomplish by adding this complexity onto the script; as i've said many times the issue you're facing is unreproducable meaning i can NOT reproduce this, in and out of a VM, tested on host (Arch Linux) and in a VM (openSUSE).
> 
> This reference
> image
> is wrong because of the !, infact it should never fail because the ! makes it so if any of them are missing it will proceed and install and since bash-completion and fontconfig aren't executables it will always run through. Along with this i suggested removing all of the deps from the command_exists function and replacing them with just BASH but of course you declined and that is the: quote me on this "solution to your nonexistent problem" as stated in the images above.
> 
> These changes do the following:
> 1. cause unnecessary overhead
> 2. accomplish the same thing as the previous version except for dep checking (which is not necessary)
> 3. potential for more unnecessary maintenance burden
> 
> please just close this non-sense, i have gave you a solution and you have declined it (for reasons unknown) this is just disgusting.

This is probably the best example of a code of conduct violation on your part. If it's not, I don't know what would qualify. You've been hostile to two contributors, myself and @lj3954, which is the only *actually* disgusting behavior on display. @ChrisTitusTech I'd like to hear your opinion on this,

cartercanedy:
This is also completely omitting any other context. Looking strictly at the course of this conversation, your behavior is completely out of line and totally antithetical to the open source process. Looking at other conversations, though, it's clear that this is a chronic issue

nnyyxxxx:
just close this, i have provided a better solution to this problem in #762

cartercanedy:
Closing in favor of #742

PR #754, opened 2024-10-04
refact: Handle temporary directories entirely within core

lj3954:
## Type of Change
- [x] Refactoring

## Description
Move all logic for handling temporary directories into linutil core. Return a struct containing the vector of tabs along with the private tempdir variable, which is held so that the destructor can be called once tabs are out of scope. Also add back temporary directory prefix ("linutil_scripts") to improve clarity.

## Impact
UIs do not need to manually hold onto a tempdir object they have no use for, improving maintainability (resolving regression in #749). 

## Checklist


adamperkowski (approved):
yep that's good

core/src/inner.rs
```diff
@@ -14,8 +15,26 @@ use temp_dir::TempDir;
 
 const TAB_DATA: Dir = include_dir!("$CARGO_MANIFEST_DIR/tabs");
 
-pub fn get_tabs(validate: bool) -> (TempDir, Vec<Tab>) {
-    let (temp_dir, tab_files) = TabList::get_tabs();
+// Allow the unused TempDir to be stored for later destructor call
+#[allow(dead_code)]
+pub struct TabList(pub Vec<Tab>, TempDir);
+
+// Implement deref to allow Vec<Tab> methods to be called on TabList
```
adamperkowski, line 22: Wait those aren't LLM generated, right?
Because I think the code here is pretty self explanatory.

lj3954: No, they aren't. Those with less familiarity with Rust may not understand what utility the deref implementation serves. The comment above the struct is also there to ensure nobody ends up removing the TempDir object from the struct, given that it's implicitly dropped and may not immediately be clear to the reader.

nnyyxxxx (approved):
nice

cartercanedy (approved):
Was thinking the same thing. Nice work

jeevithakannan2 (approved):

adamperkowski:
You've got some conflicts in `state.rs`

nnyyxxxx:
update docgen.rs as well

nnyyxxxx (approved):
amazing !!

ChrisTitusTech (approved):

PR #755, opened 2024-10-04
implement word wrapping functionality

nnyyxxxx:
## Type of Change
- [x] New feature

## Description
adds dynamic / auto updating word wrapping to the descriptions


adamperkowski (approved):
Looks OK from a first glance.

afonsofrancof:
I removed the FloatingTextMode a few days ago, so we can use any title we want.
Maybe it's better to have the new() function receive a boolean telling it if it should wrap the words, and we only activate that on the description for now.
It would then be very easy to activate it in the future for other float types

nnyyxxxx:
> > that's already implemented via enum rather than boolean
> 
> Yes, I know, that enum existed a few days ago, and it was removed to make the float more generic, so that the title can be anything instead of just those 3 enum types.
> 
> That's why I was saying that a boolean is better.

oh i didnt notice

adamperkowski (approved):

afonsofrancof (approved):
lgtm

ChrisTitusTech (approved):

PR #756, opened 2024-10-04
refact: remove snaps

jeevithakannan2:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
- Check if snap is installed before trying to remove snap.
- Merge dnf and zypper packagers as they both have the same commands.

error

no-error

## Testing
- Works as expected

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #757, opened 2024-10-04
chore: Remove unwanted case

jeevithakannan2:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
- Unsupported package manager is already handled in common-script.sh  https://github.com/ChrisTitusTech/linutil/blob/26d0adc829ac59893666559f1c126d05b7fddc6b/core/tabs/common-script.sh#L86
- Removed pacman from linutil-installer.sh cargo installation section as pacman case is already handled.
- Remove snaps script is excluded as it may conflict with #756 

## Testing
- No errors works as expected.

## Checklist


adamperkowski (approved):
NIT: I'm not sure if every script supports every "officially" supported manager, If not, we need to fix this.

jeevithakannan2:
If another package manager is added it will be handled anyway. For example, if we are adding nix then we will be adding nix to most of the scripts. If nix cannot be used for a specific script then it will be handled as unsupported. Having extra bloat that will not be reached is useless anyway.

lj3954:
> If nix cannot be used for a specific script then it will be handled as unsupported.

Yes. And this PR proposes the removal of that error handling.

jeevithakannan2:
Having extra bloat that cannot be reached be useless anyway 🤷🏻‍♂️

lj3954:
> Having extra bloat that cannot be reached is useless anyway 🤷🏻‍♂️ We can always add them back to scripts that don't support packagers later

They cannot be reached **assuming all package managers supported by linutil are supported by the script**. I've just stated a way in which that condition could be broken. Your idea of adding this code back in the case of any changes illustrates why removing it is a poor decision to begin with.

PR #758, opened 2024-10-04
do not exit linutil in confirmation prompt when q is pressed

nnyyxxxx:
## Type of Change
- [x] Bug fix

https://github.com/user-attachments/assets/e7ee90e5-fa73-4823-a284-597c361d2f25

### DIFF
--- a/tui/src/confirmation.rs
+++ b/tui/src/confirmation.rs
@@ -88,7 +88,7 @@ impl FloatContent for ConfirmPrompt {
         use KeyCode::*;
         self.status = match key.code {
             Char('y') | Char('Y') => ConfirmStatus::Confirm,
-            Char('n') | Char('N') | Esc | Char('q') | Char('Q') => ConfirmStatus::Abort,
+            Char('n') | Char('N') | Esc | Char('q') => ConfirmStatus::Abort,
             Char('j') => {
                 self.scroll_down();
                 ConfirmStatus::None
@@ -116,7 +116,7 @@ impl FloatContent for ConfirmPrompt {
             "Confirmation prompt",
             Box::new([
                 Shortcut::new("Continue", ["Y", "y"]),
-                Shortcut::new("Abort", ["N", "n", "q", "Q", "Esc"]),
+                Shortcut::new("Abort", ["N", "n", "q", "Esc"]),
                 Shortcut::new("Scroll up", ["k"]),
                 Shortcut::new("Scroll down", ["j"]),
                 Shortcut::new("Close linutil", ["CTRL-c"]),
--- a/tui/src/confirmation.rs
+++ b/tui/src/confirmation.rs
@@ -88,7 +88,7 @@ impl FloatContent for ConfirmPrompt {
         use KeyCode::*;
         self.status = match key.code {
             Char('y') | Char('Y') => ConfirmStatus::Confirm,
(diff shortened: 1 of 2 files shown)


tui/src/confirmation.rs
```diff
@@ -88,7 +88,7 @@ impl FloatContent for ConfirmPrompt {
         use KeyCode::*;
         self.status = match key.code {
             Char('y') | Char('Y') => ConfirmStatus::Confirm,
-            Char('n') | Char('N') | Esc => ConfirmStatus::Abort,
+            Char('n') | Char('N') | Esc | Char('q') | Char('Q') => ConfirmStatus::Abort,
```
adamperkowski, line 91: We have to be consistent with this. Here you're doing `Char('q') | Char('Q')` but other modules like `float.rs` or `state.rs` use `Char('q')` only.

tui/src/state.rs
```diff
@@ -411,11 +411,9 @@ impl AppState {
...
-            self.focus,
-            Focus::TabList | Focus::List | Focus::ConfirmationPrompt(_)
-        ) && (key.code == KeyCode::Char('q')
-            || key.modifiers.contains(KeyModifiers::CONTROL) && key.code == KeyCode::Char('c'))
+        if matches!(self.focus, Focus::TabList | Focus::List)
+            && (key.code == KeyCode::Char('q')
+                || key.modifiers.contains(KeyModifiers::CONTROL) && key.code == KeyCode::Char('c'))
         {
             return false;
         }
```
adamperkowski, line 419: you forgot about `Ctrl-C`

```suggestion
        }
        
        if matches!(self.focus, Focus::ConfirmationPrompt(_))
            && (key.modifiers.contains(KeyModifiers::CONTROL) && key.code == KeyCode::Char('c'))
        {
            return false;
        }
```

adamperkowski (changes requested):

adamperkowski:
cargo fmt seeing ghosts lol. wants to change `{` to `{`

adamperkowski:
9641e6474c6e73581f9ce9f846e36f9174b56808 Excuse me what do you mean untested? It literally does what it says.
image
It's not supposed to close the popup on ctrl-c, it's supposed to exit.

nnyyxxxx:
> [9641e64](https://github.com/ChrisTitusTech/linutil/commit/9641e6474c6e73581f9ce9f846e36f9174b56808) Excuse me what do you mean untested? It literally does what it says. image It's not supposed to close the popup on ctrl-c, it's supposed to exit.

ah

tui/src/state.rs
```diff
@@ -411,15 +411,23 @@ impl AppState {
...
 
+        if matches!(self.focus, Focus::ConfirmationPrompt(_))
+            && (key.modifiers.contains(KeyModifiers::CONTROL) && key.code == KeyCode::Char('c'))
+        {
+            self.focus = Focus::List;
+            if !self.multi_select {
+                self.selected_commands.clear();
+            }
+            return true;
```
adamperkowski, line 428: image

No need for `Ctrl-C` to do the same thing. 

```suggestion
            return false;
```

adamperkowski (changes requested):

adamperkowski (approved):
🍟

ChrisTitusTech (approved):

PR #774, opened 2024-10-06
fix: fedora timeshift

### LINKED ISSUE #773: [BUG] Timeshift tries and fails to place backups to /boot in Fedora.
## Describe the bug
When I tried to do backup, utility tries to save backup to /boot device and fails.

Fedora uses EXT4 for /boot and btrfs for root(/). Therefore Timeshift is not compatible with default Fedora Installs.

But even if you fix this and manage to create a backup in external drive, restore is going to fail because of btrfs root.

## To Reproduce
1. Launch Timeshift utility
2. Enter 3

## Expected behavior
Successful backup (or a warning for Fedora users?)

## Screenshots
image
image

## Additional context

I created a ackup to external disk from GUI. It gives Unsupported Subvolume layout message as I mentioned in the description.
image

## Checklist
- [.] I checked for duplicate issues.
- [.] I checked already existing discussions.
- [.] This issue is not included in the roadmap.
- [.] This issue is present on both stable and development branches.

nnyyxxxx:
### @ChrisTitusTech this does not make timeshift fedora only, this only makes it so timeshift cannot be seen by fedora users, please look at #773 for more information

closes #773 

## Type of Change
- [x] Bug fix

## Description
since fedora does not support the use of timeshift, it should be excluded.

more information is in this issue #773 

## testing
Tested using a fedora container and my main system (Arch Linux), and the precondition works flawlessly

image

### DIFF
--- a/core/tabs/utils/tab_data.toml
+++ b/core/tabs/utils/tab_data.toml
@@ -158,6 +158,11 @@ name = "Timeshift Backup"
 script = "timeshift.sh"
 task_list = "I"
 
+[[data.preconditions]]
+matches = false
+data = "command_exists"
+values = [ "dnf" ]
+
 [[data]]
 name = "WiFi Manager"
 description = "This utility is designed to manage wifi in your system"


Angxddeep:
It would be better if you also disable firewall setup for fedora, as it uses Firewalld which is on out of the fox and fedora also has rules for Workstation set out of the box, so there is really no need to set up ufw. Also, it might not work properly if both of them are on at the same time.

AkkhilCodingHub (approved):

jeevithakannan2 (approved):

adamperkowski (approved):

ChrisTitusTech (approved):

PR #789, opened 2024-10-09
feat: btrfs assistant & grub-btrfs

leventbesli:
## Type of Change
- [+] New feature
- [+] Documentation update

image

## Description
- I added a script under Fedora section to installs Btrfs Assistant with snapper and Grub-Btrfs to Fedora, allowing snapshots to be displayed in the Grub menu.

- Additionally, it configures scheduled timeline snapshots for both the root and home directories, ensures automatic GRUB updates following snapshot creation, and includes a DNF plugin that facilitates automatic snapshots before/after package installations.

- Because of default Fedora disk partitioning during OS installation, the /boot configured as an separate EXT4 partition in Fedora. Therefore, it cannot be included in root snapshots. 

Necessary information has been provided in the documentation, description and post install steps for users to pay attention to this point.
image
image

Fedora's EXT4 /boot, btrfs root(/) choice also prevents restoring Timeshift backups. So snapper still seems to be the best solution for Fedora anyway.

- The grub-btrfs installation was left optional because it is a deeper system intervention and some systems have problems with snapshot booting due to the “tpm.c:150:unknown TPM error” error. This seems like a common problem even with unmodified Fedora systems... https://bugzilla.redhat.com/show_bug.cgi?id=2128485
image

- A mitigation step was also provided for those who encountered the error.
image

## Testing
- Tested with Fedora 40 Workstation and KDE Spin without issues.
- I also did a test to see if updating the kernel after installation would cause any issues with GRUB. After the kernel was updated, GRUB also updated successfully.
- For testing purposes I completely deleted /etc and /usr directories and then did a restore to a previous snapshot  successfully.

image
Screenshot_20241003_180931
Screenshot_20241003_181007
Screenshot_20241003_181126

## Impact
Users often want to use the btrfs snapshot feature, but guides on the internet seems too complicated because they start from partitioning steps from initial setup.  I tried to perform the snapshot process in the easiest way without interfering with the partitioning.

As long as /boot is not corrupted, Fedora users can recover their systems with these tools. They may need to evaluate manual backup options for /boot.

## Checklist
- [+] My code adheres to the coding and style guidelines of the project.
- [+] I have performed a self-review of my own code.
- [+] I have commented my code, particularly in hard-to-understand areas.
- [+] I have made corresponding changes to the documentation.
- [+] My changes generate no errors/warnings/merge conflicts.


leventbesli:
Re-opened PR with docgen updated. Thanks for useful tool @adamperkowski

adamperkowski (approved):
> Re-opened PR with docgen updated. Thanks for useful tool @adamperkowski

Thanks. Was merged in #784.

ChrisTitusTech (approved):

PR #821, opened 2024-10-13
Add root check

### LINKED ISSUE #816: remove "cant find escaltion tool if user is already root
## Describe the bug
log in as root user without a escation tool, eg docker

## To Reproduce
run the script
select an opetion which requires elevation
## Expected behavior
the script to run normal

## Checklist
- [*] I checked for duplicate issues.
- [*] I checked already existing discussions.
- [*] This issue is not included in the roadmap.
- [*] This issue is present on both stable and development branches.

jeevithakannan2:
# Only merge this at last after merging all the script PRs
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [x] Security patch
- [ ] UI/UX improvement

- **Thanks @cartercanedy @adamperkowski @lj3954 for helping out !!**
- Add root check and warn users about running as root
- Uses `nix` crate

image

## Testing
- Tested with no issues !!

## Issues / other PRs related
- Resolves #816 

## Checklist


solomoncyj:
what if i am in container system such as docker/podman? they do not drop me as a normal user with dmin rights. they just drop the user in as the root acc.

solomoncyj:
just look for `.dockerenv` in /

jeevithakannan2:
> just look for `.dockerenv` in /

Not only docker env runs as root. Normal user may invoke linutil as root

solomoncyj:
> > just look for `.dockerenv` in /
> 
> Not only docker env runs as root. Normal user may invoke linutil as root

this allows the user run the script to run it in docker/podman/ andy other container system. this is because users do not usally create users in containers, and insted default to the root user pre-created. we just need to tewak the scripts to allow the scripts to skip elevation checks if the user is root

jeevithakannan2:
> > > just look for `.dockerenv` in /
> > 
> > 
> > Not only docker env runs as root. Normal user may invoke linutil as root
> 
> this allows the user run the script to run it in docker/podman/ andy other container system. this is because users do not usally create users in containers, and insted default to the root user pre-created. we just need to tweak the scripts to allow the scripts to skip elevation checks if the user is root

Running scripts as root may break many things. For example configuration files should not be in root permission. An application or non-root user may not be able to rw the configuration file due to insufficient permissions. And why would someone try to run linutil in a docker container ??

solomoncyj:
I need to set up the bash scripts as I can't stand the default one

cartercanedy:
@jeevithakannan2 instead of precluding the use, it might be worth implementing a root warning and doing a refactor to just leave `ESCALATION_TOOL` unset when run as root. Configuration files *should* written relative to $HOME, which would be inaccessible to other users anyways, and software is almost exclusively, if not completely, handled by package managers, so there shouldn't be any permission issues there. If there *are* any game-breaking conflicts, I'll eat my words, but I don't think that we need to put that tight of a restriction.

cartercanedy:
> just look for `.dockerenv` in /

I don't like this. Running a ubuntu docker container should generally the same as running as root in a bare-metal ubuntu install, unless you're running a crazy stripped-down image. There are better solutions to make this workable for this kind of use-case

jeevithakannan2:
> @jeevithakannan2 instead of precluding the use, it might be worth implementing a root warning and doing a refactor to just leave `ESCALATION_TOOL` unset when run as root. Configuration files *should* written relative to $HOME, which would be inaccessible to other users anyways, and software is almost exclusively, if not completely, handled by package managers, so there shouldn't be any permission issues there. If there *are* any game-breaking conflicts, I'll eat my words, but I don't think that we need to put that tight of a restriction.

Ok will go through this one more time. If root is disabled arch server setup cannot be run. I think I should do it in the scripts then

cartercanedy:
Also, I made a PR on your dev branch that implements a warning prompt notifying users that they are running as root. Since we're not really testing these scripts with standard user perms, it's probably best to notify users that they are outside the bounds of the current user story

cartercanedy:
I also would remove the cli flag since most users are probably going to be using the network runner

jeevithakannan2:
> Also, I made a PR ([jeevithakannan2#1](https://github.com/jeevithakannan2/linutil/pull/1)) on your dev branch that implements a warning prompt notifying users that they are running as root. Since we're not really testing these scripts with standard user perms, it's probably best to notify users that they are outside the bounds of the current user story

Can't we just add a flag

> I also would remove the cli flag since most users are probably going to be using the network runner

Yeah removed it thanks for the PR

adamperkowski (changes requested):
I think this implementation is wrong and overengineered. We can just check the UID or even file permissions. This will only slow linutil down.

cartercanedy:
> This will only slow linutil down

?

adamperkowski:
> > This will only slow linutil down
> 
> ?

This PR is adding another rust dependency, more checks and code that has to be maintained. I just think this can be easily avoided with a more efficient approach.

cartercanedy:
To get the inlined superuser check functional would need even more code to maintain, and it would be hard to beat 8 lines of bash except by just unquoting all references to`ESCALATION_TOOL` and not bothering with an additional function

cartercanedy:
Tbh, I think unquoting would be the simplest @jeevithakannan2

core/tabs/common-script.sh
```diff
@@ -44,6 +52,12 @@ checkAURHelper() {
 
 checkEscalationTool() {
     ## Check for escalation tools.
+    if [ "$(id -u)" -eq 0 ]; then
+        printf "%b\n" "${CYAN}Running as root, no need for escalation tool.${RC}"
```
adamperkowski, line 56:
```suggestion
    if [ "$(id -u)" -eq 0 ]; then
        printf "%b\n" "${CYAN}Running as root. Skipping escalation tool check.${RC}"
```

tui/src/root.rs
```diff
@@ -0,0 +1,69 @@
...
+
+pub fn check_root(terminal: &mut Terminal<CrosstermBackend<io::Stdout>>) -> io::Result<bool> {
+    if sudo::check() == sudo::RunningAs::Root {
+        terminal.draw(|frame| {
+            let root_warn = Paragraph::new(
+                r#"
+!!!!!!!!!!!!!! YOU ARE ABOUT TO RUN LINUTIL AS ROOT !!!!!!!!!!!!!!
+This utility prioritizes compatibility with non-root environments.
+Some scripts may work without any issues, some may not.
+You have been warned!
```
adamperkowski, line 19: i think we have enough exclamation points

```suggestion
You have been warned.
```

jeevithakannan2:
Mmm 🤔

adamperkowski:
@cartercanedy To be 100% honest, I love this idea
BUT
depending on a crate that hasn't been updated in 3-4 years for **privilage-escalation** related stuff is just not a good idea.
image

adamperkowski:
> Getting the inlined superuser check functional would need even more code to maintain, and it would be hard to beat 8 lines of bash except by just unquoting all references to`ESCALATION_TOOL` and not bothering with an additional function

I'm talking about Rust code.

core/tabs/applications-setup/Developer-tools/ngrok-setup.sh
```diff
@@ -5,12 +5,18 @@
...
     if ! command_exists ngrok; then
         printf "%b\n" "${YELLOW}Installing Ngrok...${RC}"
-        curl -sSLO https://bin.equinox.io/c/bNyj1mQVY4c/ngrok-v3-stable-linux-amd64.tgz | "$ESCALATION_TOOL" tar -xz -C /usr/local/bin
+        curl -sSLO https://bin.equinox.io/c/bNyj1mQVY4c/ngrok-v3-stable-linux-amd64.tgz
+        if [ -f "ngrok-v3-stable-linux-amd64.tgz" ]; then
+            elevated_execution tar -xz -f ngrok-v3-stable-linux-amd64.tgz -C /usr/local/bin
+        else
+            printf "%b\n" "${RED}Error occurred when downloading.${RC}"
+            exit 1
+        fi
```
adamperkowski, line 14: No need for this check. The script will stop when any command (including `curl`) fails because of `sh -e`.

adamperkowski (changes requested):

tui/src/root.rs
cartercanedy: Lmao

jeevithakannan2:
So I undo the changes and not use a crate. If yes I have solution for that too !!

cartercanedy:
> depending on a crate that hasn't been updated in 3-4 years for **privilege-escalation** related stuff is just not a good idea.

Fair point

jeevithakannan2:
We can check the USER env right for running as root ?

core/tabs/applications-setup/Developer-tools/ngrok-setup.sh
jeevithakannan2: The if condition is not necessary but piping was causing issues so I split the command

adamperkowski:
@jeevithakannan2 You can use Rust's `libc` to check current UID.

jeevithakannan2:
> `libc`

Yeah thought of that. It needs unsafe that's why I leaned toward crate

adamperkowski:
Yeah we currently 0% of Linutil's codebase is unsafe. Would be sad to loose that.

adamperkowski:
What about `nix::unistd`?

cartercanedy:
> How about `nix::unistd`?

That seems like the best safe option

jeevithakannan2:
Yeah agreed

jeevithakannan2:
What about the scripts ??

adamperkowski:
> What about the scripts ??

scripts are fine

core/tabs/applications-setup/Developer-tools/ngrok-setup.sh
adamperkowski: okay

jeevithakannan2:
> > What about the scripts ??
> 
> scripts are fine

We can just unquote the ESCALATION_TOOL instead of creating a new function

tui/src/root.rs
jeevithakannan2: Exclamation looks good

lj3954:
I'm not sure what all this discussion is, it's already been decided that we're going to adapt the scripts to work on root (as as been done in the recent commits here) rather than prohibiting linutil from launching as root. Please drop https://github.com/ChrisTitusTech/linutil/pull/821/commits/c997a3d89e6908a6fb76d15da03ee43a6ed4fdda. 

By the way, @adamperkowski, both crates mentioned here, nix and this unmaintained sudo crate use libc::getuid() anyways. I think it's a poor idea to religiously avoid writing a line of unsafe code, just to pick up another dependency which calls quite literally the exact same function. That does the opposite of improving security.

adamperkowski:
> By the way, @adamperkowski, both crates mentioned here, nix and this unmaintained sudo crate use libc::getuid() anyways. I think it's a poor idea to religiously avoid writing a line of unsafe code, just to pick up another dependency which calls quite literally the exact same function. That does the opposite of improving security.

I am aware that they do use `libc::getuid` but by not including unsafe code blocks in our code directly we at least know that future PRs / other stuff like that will be relatively safe and have an overall lower chance of causing something like memory leaks, overflows, etc. It is 100% possible that (for example) `libc` gets updated in the future and using `getuid` directly could cause issues. Using `nix::unistd` solves this issue. (assuming that `nix` would be regularly updated)

cartercanedy:
We should use a `#![forbid(unsafe_code)]` top-level directive to enforce safety then

adamperkowski (approved):

cartercanedy (approved):

lj3954 (approved):

jeevithakannan2:
@ChrisTitusTech All the scripts have shellcheck warning and formatting issues that need to be resolve in a separate PR. This PR has no issues.

adamperkowski:
?

cartercanedy:
> ?

I was just about to say...

PR #699, opened 2024-09-29
feat(bootable-usb): add Fedora

lennartpj:
## Type of Change
- [x] New feature
- [x] Bug fix

## Description
This pull request adds functionality to download the latest Fedora 40 Workstation ISO in the "create-bootable-usb.sh". It also fixes a bug in the confirmation prompt, allowing users to proceed by entering 'y', 'yes', and case-insensitive variants.

## Testing
I have tested the changes by executing cargo run and going through the steps. 
After the Script executed I was able to boot from the USB stick into the Fedora Setup.

The confirmation prompt now accepts 'y', 'yes', and case-insensitive variants.

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #828, opened 2024-10-15
feat: Drastically increase the operating system support for USB creation script

lj3954:
## Type of Change
- [x] New feature

## Description
Remove inbuilt logic for finding debian & arch ISO URLs from the USB creation script. Instead, if the user selects the online ISO option, download JSON data containing the necessary information from a project I started and brought into the quickemu project, [quickget_configs](https://github.com/quickemu-project/quickget_configs). This project was created to eventually be used to fully replace functionality in the quickget bash script, similar to that present in this script. Thus, it provides the data necessary to download images of various operating systems. The data is produced daily in CI, where invalid or unresolvable URLs (and their accompanying entries) are thrown out, and other useful information, like SHA256 checksums to verify validity, are included. 

jq is used to filter through this JSON data, and ensure that we're only taking entries with 1 ISO file and nothing else.  The user is prompted for the Architecture, OS, Release, and Edition (if necessary), and the ISO is downloaded, verified, and unarchived as applicable.  This change will additionally reduce the maintenance burden on this script, since the logic to find URLs is contained within a different project.

I also found and removed an unused usage function. I suppose someone forgot to remove that after asking an LLM to write an entire script for them... @guruswarupa 

https://github.com/user-attachments/assets/132f27af-28a9-4c1b-9cba-71cba7cdd5ca

## Impact
jq is added as a dependency for the script, along with a few archiving tools (xz, bzip2, gzip).

## Issues / other PRs related

- Functionality supersedes that in #699. 
- New installDependencies function would have to be updated on event of #821 merge.

## Checklist

### DIFF
--- a/core/tabs/utils/create-bootable-usb.sh
+++ b/core/tabs/utils/create-bootable-usb.sh
@@ -2,12 +2,7 @@
 
 . ../common-script.sh  
 
-# Function to display usage instructions
-usage() {
-    printf "%b\n" "${RED} Usage: $0 ${RC}"
-    printf "%b\n" "No arguments needed. The script will prompt for ISO path and USB device."
-    exit 1
-}
+CONFIGURATION_URL="https://github.com/quickemu-project/quickget_configs/releases/download/daily/quickget_data.json"
 
 # Function to display all available block devices
 list_devices() {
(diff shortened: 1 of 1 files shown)


lj3954:
Older jq versions appear to break in many cases here. Working to resolve it

nnyyxxxx (approved):
nice changes

core/tabs/utils/create-bootable-usb.sh
```diff
@@ -94,43 +59,151 @@ choose_iso_source() {
...
             ;;
     esac
+}
 
-    ISO_PATH=$(basename "$ISO_URL")
-    printf "%b\n" "${YELLOW}Downloading ISO...${RC}"
-    curl -L -o "$ISO_PATH" "$ISO_URL"
-    if [ $? -ne 0 ]; then
-        printf "%b\n" "${RED}Failed to download the ISO file.${RC}"
+# Function to fetch ISO URLs
```
nnyyxxxx, line 128:
```suggestion
```
this comment is unnecessary

nnyyxxxx (changes requested):

core/tabs/utils/create-bootable-usb.sh
nnyyxxxx:
```bash
get_online_iso() {
    get_architecture
    printf "%b\n" "${YELLOW}Fetching available operating systems...${RC}"
```
the purpose is described perfectly here

lj3954: Indeed. That comment was there prior to my changes, git just tagged it as a line removed & added.

core/tabs/utils/create-bootable-usb.sh
```diff
@@ -2,12 +2,7 @@
 
 . ../common-script.sh  
 
-# Function to display usage instructions
-usage() {
-    printf "%b\n" "${RED} Usage: $0 ${RC}"
-    printf "%b\n" "No arguments needed. The script will prompt for ISO path and USB device."
-    exit 1
-}
+CONFIGURATION_URL="https://github.com/quickemu-project/quickget_configs/releases/download/daily/quickget_data.json"
```
nnyyxxxx, line 5: can we not externally link projects that could **potentially** be **dangerous**, this is a risk i'm not sure Chris would like to take, maybe we should host this in this repository so it can be monitored.

nnyyxxxx (changes requested):

core/tabs/utils/create-bootable-usb.sh
lj3954: Please read the body of my PR. That is quite literally the entire proposal of the PR. The project that is being relied upon here is extremely trustworthy. The Quickemu project was created by Martin Wimpress, who is an extremely trusted member of the community. He's a former Canonical employee and long term lead of Ubuntu MATE. In fact, Chris has used and showcased quickemu's bash scripts & flutter GUI in the past (https://www.youtube.com/watch?v=87MCT3Ekjlo). 

The quickget_configs tool, in specific, was created by me. I had several discussions with other contributors to the quickemu project along with the lead, and the idea of creating a tool which generates all the data necessary to create a QEMU virtual machine, including URLs to download image files, in CI was brought up. I worked on it for a bit alongside my re-implementations of the project's 2 main tools, and then relinquished my ownership of the project to Martin's quickemu project. Once again, it's in very safe hands. The project will not be hosted in this repository, for reasons that should already be spelled out well enough.

It's also worth looking at what a potential attack on the project could even affect. Someone would have to insert a mirror providing malicious images into the logic for creating configurations to download one operating system. There should and will be high scrutiny on unrecognized URLs in future contributions to the project, and the reward for the attacker for carrying out such a complicated attack would be miniscule (it would only affect people who chose to download that one specific operating system, nearly exclusively for virtual machines). If we are to be worried about a security risk, a MITM attack (which this change would protect against, due to the checksums being fetched in CI) or a legitimate mirror being compromised are **far** more likely threats.

nnyyxxxx (approved):

core/tabs/utils/create-bootable-usb.sh
```diff
@@ -94,43 +59,150 @@ choose_iso_source() {
...
+
+    OS_JSON="$(echo "${OS_JSON}" | jq --arg os "${OS}" -c '.[] | select(.name == $os)')"
+    if [ -z "${OS_JSON}" ]; then
+        printf "%b\n" "${RED}Invalid operating system selected.${RC}"
+        exit 1
+    fi
+    PRETTY_NAME="$(echo "${OS_JSON}" | jq -r '.pretty_name')"
+
+    printf "\n%b\n" "${YELLOW}Available releases for ${PRETTY_NAME}:${RC}"
+    echo "${OS_JSON}" | jq -r '.releases[].release' | sort -Vur | tr '\n' ' '
```
adamperkowski, line 160:
```suggestion
    echo "${OS_JSON}" | jq -r '.releases[].release' | sort -Vur | tr '\n' ';'
```

core/tabs/utils/create-bootable-usb.sh
```diff
@@ -94,43 +59,150 @@ choose_iso_source() {
...
+        and .img == null and .fixed_iso == null and .floppy == null and .disk_images == null)) \
+        | select(.releases | length > 0)]")"
+
+    if echo "${OS_JSON}" | jq -e 'length == 0' >/dev/null; then
+        printf "%b\n" "${RED}No operating systems found.${RC}"
         exit 1
     fi
+
+    printf "%b\n" "${YELLOW}Available Operating Systems:${RC}"
+    echo "${OS_JSON}" | jq -r '.[].name' | tr '\n' ' '
```
adamperkowski, line 148:
```suggestion
    echo "${OS_JSON}" | jq -r '.[].name' | tr '\n' ';'
```

adamperkowski (approved):

ChrisTitusTech (approved):

PR #856, opened 2024-10-20
fix+refact(system-cleanup)

### LINKED ISSUE #855: Full System Cleanup doesn't clean Trash if it errors out on cleaning cache
## Describe the bug
Full System Cleanup doesn't perform all tasks because one of these independent tasks has failed. When cleaning the cache can't run (in this case because .cache doesn't exist), the script ends and doesn't run Trash cleanup

## To Reproduce
1. Disable/Stop any processes that write to .cache (including Starship)
2. Remove `~/.cache`
3. Run System Disk Cleanup from Linutil
4. Select y to clean up Cache and Trash
5. Script fails because .cache does not exist
6. Trash remains uncleaned

## Expected behavior
Script still runs Trash cleaning regardless of the failed cache-cleanup.

## Screenshots

screenshot-2024-10-20-12-28-25

screenshot-2024-10-20-12-28-47

## Additional context
Distro: Fedora 40. 
Happens with regular linutil and the dev branch too (linux and linuxdev)

## Checklist

adamperkowski:
## Type of Change
- [x] Bug fix
- [x] Refactoring
- [x] UI/UX improvement

## Testing

Works :)

## Issues / other PRs related

- Resolves #855

## Checklist

### DIFF
--- a/core/tabs/system-setup/system-cleanup.sh
+++ b/core/tabs/system-setup/system-cleanup.sh
@@ -26,16 +26,21 @@ cleanup_system() {
             "$ESCALATION_TOOL" "$PACKAGER" -Rns $(pacman -Qtdq) --noconfirm > /dev/null 2>&1
             ;;
         *)
-            printf "%b\n" "${RED}Unsupported package manager: ""$PACKAGER""${RC}"
-            return 1
+            printf "%b\n" "${RED}Unsupported package manager: ""$PACKAGER"". Skipping.${RC}"
             ;;
     esac
 }
 
 common_cleanup() {
-    "$ESCALATION_TOOL" find /var/tmp -type f -atime +5 -delete
-    "$ESCALATION_TOOL" find /tmp -type f -atime +5 -delete
-    "$ESCALATION_TOOL" find /var/log -type f -name "*.log" -exec truncate -s 0 {} \;
+    if [ -d /var/tmp ]; then
+        "$ESCALATION_TOOL" find /var/tmp -type f -atime +5 -delete
+    fi
+    if [ -d /tmp ]; then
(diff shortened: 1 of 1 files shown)


core/tabs/system-setup/system-cleanup.sh
```diff
@@ -26,16 +26,21 @@ cleanup_system() {
...
-    "$ESCALATION_TOOL" find /var/log -type f -name "*.log" -exec truncate -s 0 {} \;
+    if [ -d /var/tmp ]; then
+        "$ESCALATION_TOOL" find /var/tmp -type f -atime +5 -delete
+    fi
+    if [ -d /tmp ]; then
+        "$ESCALATION_TOOL" find /tmp -type f -atime +5 -delete
+    fi
+    if [ -d /var/log ]; then
+        "$ESCALATION_TOOL" find /var/log -type f -name "*.log" -exec truncate -s 0 {} \;
+    fi
```
jeevithakannan2, line 43: Can we do something like this ?

```suggestion
    "$ESCALATION_TOOL" find /var/tmp -type f -atime +5 -delete &> /dev/null || true
    "$ESCALATION_TOOL" find /tmp -type f -atime +5 -delete &> || true
    "$ESCALATION_TOOL" find /var/log -type f -name "*.log" -exec truncate -s 0 {} \; &> || true
```

adamperkowski: We could but if user doesnt have perms for a file (could happen and will happen with /tmp) it'll fail too

jeevithakannan2: The cmd is escalated so there will no perm err

ChrisTitusTech (approved):

PR #884, opened 2024-10-28
fix: github release notes not working

adamperkowski:
so apparently you can't mix stuff like `labels` and `label` in YAML (f YAML)
if i try to create a release now (yes i can), it just won't generate the notes. this will fix it
image

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,23 +1,35 @@
 changelog:
   categories:
     - title: '🚀 Features'
-      label: 'enhancement'
+      labels:
+        - 'enhancement'
     - title: '🐛 Bug Fixes'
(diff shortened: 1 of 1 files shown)


.github/release.yml
```diff
@@ -1,23 +1,35 @@
...
-      label: 'rust'
+      labels:
+        - 'rust'
     - title: '📃 Scripting'
-      label: 'script'
+      labels:
+        - 'script'
+    - title: 'Other Changes'
+      labels:
+        - "*"
```
adamperkowski, line 32: Oh and also we won't have to label **every single** PR with this.

ChrisTitusTech (approved):

PR #913, opened 2024-11-04
feat: Mouse scroll

jeevithakannan2:
## Type of Change
- [x] New feature
- [x] UI/UX improvement

## Description
- Add mouse scroll in all directions.

https://github.com/user-attachments/assets/d666c77f-37ec-4469-8ff7-130d14da4e23

## Testing
- Tested on arch linux with no issues.

## Checklist


tui/src/main.rs
```diff
@@ -70,15 +72,23 @@ fn run(
...
-                return Ok(());
+                if !state.handle_key(&key) {
+                    return Ok(());
+                }
+            }
+            Event::Mouse(mouse_event) => {
+                if !state.handle_mouse(&mouse_event) {
+                    return Ok(());
+                }
+                continue;
```
lj3954, line 89:
```suggestion
```
No need for this.

tui/src/confirmation.rs
```diff
@@ -84,6 +84,21 @@ impl FloatContent for ConfirmPrompt {
...
+            MouseEventKind::ScrollDown => {
+                self.scroll_down();
+                ConfirmStatus::None
+            }
+            MouseEventKind::ScrollUp => {
+                self.scroll_up();
+                ConfirmStatus::None
+            }
+            _ => ConfirmStatus::None,
+        };
```
lj3954, line 98:
```suggestion
        match event.kind {
            MouseEventKind::ScrollDown => {
                self.scroll_down();
            }
            MouseEventKind::ScrollUp => {
                self.scroll_up();
            }
            _ => {}
        }
```
Why is this match statement returning an unused ConfirmStatus?

tui/src/state.rs
```diff
@@ -416,6 +416,45 @@ impl AppState {
...
+            Focus::FloatingWindow(float) => {
+                float.content.handle_mouse_event(event);
+            }
+            Focus::ConfirmationPrompt(confirm) => {
+                confirm.content.handle_mouse_event(event);
+            }
+            _ => {}
+        }
+        true
+    }
```
lj3954, line 456: We should consider taking the position of the mouse into account. For example, mouse events outside a floating window should be ignored, and the tablist/list focus should switch if a mouse event occurs in the other.

lj3954 (changes requested):

tui/src/confirmation.rs
jeevithakannan2: I was wondering the same with handle_key_event

tui/src/state.rs
jeevithakannan2: Yeah I get it. That was my idea but it requires little bit more time. I have my semesters from next week will need to do that after my semesters. For now this is okayish

jeevithakannan2: Implemented !!

nnyyxxxx (changes requested):
this should be togglable and off by default

jeevithakannan2:
> this should be togglable and off by default

Why would someone want that?

nnyyxxxx:
> > this should be togglable and off by default
> 
> Why would someone want that?

this is a tui application ran in the terminal having mouse interaction without a toggle is poor ux by default

ChrisTitusTech (approved):

Issue #932, opened 2024-11-10
Search text disappears if arrow keys are used


jeevithakannan2:
## Describe the bug
- The text preview in search bar is bugged

## To Reproduce
1. Enter Search Mode
2. Search anything make sure the suggestion appears
3. Use arrow keys to move front and back
4. See the bug

## Expected behavior
The preview should not hide the text after the cursor.

## Screenshots

Here the word alacritt is typed. The ty appears as a suggestion while hiding the words ac.

![image](https://github.com/user-attachments/assets/0883d29c-6491-4390-8ddc-bec56be3f963)

https://github.com/user-attachments/assets/5bb32b42-89e7-4923-8ced-c90125a2bfb7

## Checklist
- [x] I checked for duplicate issues.
- [x] I checked already existing discussions.
- [x] This issue is not included in the roadmap.
- [x] This issue is present on both stable and development branches.

PR #933, opened 2024-11-10
refact: rust fixes and optimizations

jeevithakannan2:
Thanks @lj3954 and @cartercanedy for your reviews and suggestions. Learnt a lot from you guys 😄

## Type of Change
- [x] Bug fix
- [x] Refactoring
- [x] UI/UX improvement

## Notable changes so for
- 1aa1dfe3823c1534aab93164d17f9619c66dd445 Use specified theme color instead of using `ratatui::Color` for `running_command.rs` success and fail color + `filter.rs` preview text color + `state.rs` min-tui warning color, `confirmation.rs` add colors for confirmation prompt, `theme.rs` fix inverted success and fail colors 
- 190c26cd76bd535ac88231ed61d157dea33aacbe Change `ConfirmPrompt` struct fields to `private`. Prevent scrolling beyond the list. Fix color bleeding.
- 0b4f33c761f27a2efb12628b60cb6b6afc89cd28 Use `sort_unstable_by` no need of `sort_by` here, fix word disappearing by recalculating the render x for preview text.
- 10352c62549f4a4e492759c360b8796a7b98a1f8 Replace `anstyle` and `text-wrap` with `ratatui` inbuilt functions. Prevent processing of text every frame by caching the processed text.
~~- 6e5fe07fe351b2a98f5463ec661145758562477d Adds back the removed clap features from cargo toml.~~ https://github.com/ChrisTitusTech/linutil/pull/935
- 3d0f7f4c13e7395e0a8b8fbb608e2065ca12c665 Reference instead of borrowing commands from state, Refactor draw function variables to immutable, calculate inner size from block instead of manual definition.
- 79aae9eb24da0a3977d725aa5c61296ed83c4d01 Move tips to a seperate file for modularity. Pass the whole args to state instead of seperate args. Add the `longest_tab_length` to appstate struct so that it will not be calculated for each frame rendered. Use function for spawning confirmprompt. Merge command list and task items list rendering a single widget instead of two. Remove redundant keys in handle_key. 
- 1b02fc3bd84ae9cddefcbc4c7371faef638a910f Use `const` for theme functions.
- 1044517d8e26183c8ee3224d588b4878109feafb Prevent getting locked out when in floating window.
- 78b7c3faa1aa89ee86ab9a30380083789c693e18 Add more options to the config files

## Testing
- Tested with no issues.

#### Before
image

#### After
image

#### Before
https://github.com/user-attachments/assets/5bb32b42-89e7-4923-8ced-c90125a2bfb7
#### After
https://github.com/user-attachments/assets/51efc93c-247a-42a2-8a97-558cd9cc1da4

## Impact
Performance improvements and clean code base.

## Issues / other PRs related
- Resolves https://github.com/ChrisTitusTech/linutil/issues/932
- Resolves https://github.com/ChrisTitusTech/linutil/issues/953
- Closes #954

## Checklist


lj3954:
> * [0b4f33c](https://github.com/ChrisTitusTech/linutil/commit/0b4f33c761f27a2efb12628b60cb6b6afc89cd28) Use regex for case insensitive search, implement `String` instead of `char<Vec>` for `seach_input`, fix word disappearing by recalculating the render x for preview text.

Other than the last point, why would we want to do any of these things? Compiling a regex on each change to the search input makes the code less readable & negatively impacts performance compared to just taking the lowercase of the input, which we already do. Vec<char> is used to fix crashes upon inputting a multi-byte unicode character, unicode_width is used to fix incorrect cursor positioning with wider unicode characters. Both are niche issues, but I don't see the purpose of removing these fixes for absolutely no improvement.

> https://github.com/ChrisTitusTech/linutil/commit/3b7e859af80bfa1f453928a74c47efd897e26934 Replace redundant match statements with helper function in theme.rs.

Adding a new theme which has different colours for these theme elements could require these match statements be added back. This is the same thing you did when you suggested removing "unnecessary cases" in scripts. This change also does not improve readability.

I oppose these 2 commits. I'll review certain elements of the other changes in a few minutes.

tui/src/confirmation.rs
```diff
@@ -81,26 +83,25 @@ impl FloatContent for ConfirmPrompt {
             })
             .collect::<Text>();
 
-        frame.render_widget(Clear, inner_area);
         frame.render_widget(List::new(paths_text), inner_area);
     }
 
     fn handle_key_event(&mut self, key: &KeyEvent) -> bool {
-        use KeyCode::*;
+        use ConfirmStatus::*;
```
lj3954, line 90: I personally think this decreases readability, especially when `None` is one of the enum variants, easily confusable with Option::None. 

I agree with the change to KeyCode imports on the next line; it's generally not good practice to wildcard import enums with unit variants, as the removal of any unit variant you're referencing in a match statement will cause that pattern to match any variant.

tui/src/confirmation.rs
```diff
@@ -16,35 +14,30 @@ pub enum ConfirmStatus {
...
+    inner_area_height: usize,
+    names: Box<[String]>,
     scroll: usize,
+    pub status: ConfirmStatus,
 }
 
 impl ConfirmPrompt {
-    pub fn new(names: &[&str]) -> Self {
-        let max_count_str = format!("{}", names.len());
+    pub fn new(names: Vec<&str>) -> Self {
```
lj3954, line 24: This change should be reverted, it's **always** good practice to only take a reference if you don't need an owned value, whether or not your code currently passes in only owned values.

tui/src/confirmation.rs
```diff
@@ -16,35 +14,30 @@ pub enum ConfirmStatus {
...
         let names = names
             .iter()
             .zip(1..)
-            .map(|(name, n)| {
-                let count_str = format!("{n}");
-                let space_str = (0..(max_count_str.len() - count_str.len()))
-                    .map(|_| ' ')
-                    .collect::<String>();
-                format!("{space_str}{n}. {name}")
-            })
```
lj3954, line 36: This code appears to be used to make all entries start at the same position on screen in the case that you have over 10 entries.
i.e.
```
  1. value1...
 10. value10...
100. value100
```
I don't personally care one way or the other whether this remains, but I figured I'd just specify its purpose in case you didn't understand when you made the change.

nnyyxxxx:
i personally think its better to keep it lowercase as it feels more snappier that way imo

tui/src/confirmation.rs
jeevithakannan2: I think 
```
1. value1
10. value2
100. value3
```

image

Is not a big difference from if we see in the image. Maybe this is upto everyones personal preference

```
    1. value1
  10. value 2
110. value 3
```

image

tui/src/confirmation.rs
jeevithakannan2: I agree.

tui/src/confirmation.rs
jeevithakannan2: Since self.status expects ConfirmPrompt Enum think will not be an issue. I too think this decreases readability but in this context we are not using Options anywhere so user would know it is a ConfirmPrompt enum variant.

jeevithakannan2:
> > * [0b4f33c](https://github.com/ChrisTitusTech/linutil/commit/0b4f33c761f27a2efb12628b60cb6b6afc89cd28) Use regex for case insensitive search, implement `String` instead of `char<Vec>` for `seach_input`, fix word disappearing by recalculating the render x for preview text.
> 
> Other than the last point, why would we want to do any of these things? Compiling a regex on each change to the search input makes the code less readable & negatively impacts performance compared to just taking the lowercase of the input, which we already do. `Vec<char>`, over String, is used to fix panics upon inputting a multi-byte unicode character, unicode_width is used to fix incorrect cursor positioning with wider unicode characters. Both are niche issues, but I don't see the purpose of removing these fixes for absolutely no improvement.
> 
> > [3b7e859](https://github.com/ChrisTitusTech/linutil/commit/3b7e859af80bfa1f453928a74c47efd897e26934) Replace redundant match statements with helper function in theme.rs.
> 
> ~Adding a new theme which has different colours for these theme elements could require these match statements be added back. This is the same thing you did when you suggested removing "unnecessary cases" in scripts. This change also does not improve readability.~
> 
> Sorry, I misunderstood the changes made in that commit. It's even worse an idea than what I thought you were doing. If more themes are added, you'd have to keep adding function parameters, making the code even more difficult to read. A match statement is by far the best choice there.
> 
> I oppose these 2 commits. I'll review certain elements of the other changes in a few minutes.

Thanks for the quick review. I forgot about the multi-byte characters reverted back to Vec<char> with some comments.

jeevithakannan2:
@cartercanedy @lj3954 Why are using ansi colors for tree-sitter-highlighting instead of ratatui colors?

cartercanedy:
treesitter-highlight provides the machinery for interacting with the individual parsers for syntax highlighting, in our case it's the bash parser. This is necessary

Anstlye makes it easier to render the ansi color codes directly to a stream. There's some complication with using the ratatui text constructs since there might be some styling that needs to be applied over a line break, which would make parsing more complicated than necessary

tui/src/confirmation.rs
cartercanedy: This was a sticking point. I would leave it in

jeevithakannan2:
Since we using custom colors for syntax_highlighting_styles maybe using the ratatui colors reduces the complexity?

cartercanedy:
Trust me, it's just going to get more complicated...

jeevithakannan2:
Ok I will try my best and create a commit let me know what you think about it

tui/src/hint.rs
```diff
@@ -64,7 +62,7 @@ impl Shortcut {
         Self {
             key_sequences: key_sequences
                 .iter()
-                .map(|s| Span::styled(Cow::<'static, str>::Borrowed(s), Style::default().bold()))
+                .map(|s| Span::styled(*s, Style::default().bold()))
```
cartercanedy, line 65:
```suggestion
                .map(|s| Span::styled(Cow::<'_, str>::Borrowed(s) Style::default().bold()))
```

This isn't an optimization. There's no reason to deref the str slice, just use a `Cow` to prevent unnecessary copies. It's going to use a `Cow` under the hood, anyways. `Cow` is a part of `std`, so there's no benefit to removing the import either.

tui/src/hint.rs
```diff
@@ -1,5 +1,3 @@
-use std::borrow::Cow;
```
cartercanedy, line 1: Moo

cartercanedy (changes requested):

tui/src/hint.rs
jeevithakannan2: Ahh I see.

tui/src/hint.rs
jeevithakannan2: Dropped the commit Mooooo

jeevithakannan2:
Is it normal for the checkout process to take in 10min in typos workflow 🙄 https://github.com/ChrisTitusTech/linutil/actions/runs/11782804516/job/32818643344?pr=933

cartercanedy:
The initial checkout can get throttled, I've seen that happen in my own projects

jeevithakannan2:
> The initial checkout can get throttled, I've seen that happen in my own projects

Are the changes to the anstyle good ? Did you review the changes?

cartercanedy:
Lgtm

tui/src/state.rs
```diff
@@ -877,13 +877,13 @@ impl AppState {
...
+        let commands: Vec<&Command> = self
             .selected_commands
             .iter()
-            .map(|node| node.command.clone())
+            .map(|node| &node.command)
             .collect();
 
-        let command = RunningCommand::new(commands);
+        let command = RunningCommand::new(&commands);
         self.spawn_float(command, 80, 80);
```
cartercanedy, line 887: I like what I see here, nice work

cartercanedy:
I've noticed that Chris can get overwhelmed and slightly annoyed with larger PRs, and I think this is starting to become too sprawling. I would break it down so that each PR is more atomic and understandable in its own context.

jeevithakannan2:
> I don't want to put words in his mouth, but I've noticed that Chris can get overwhelmed and slightly annoyed with larger PRs. With that said, I think this is starting to become too sprawling; I would break it down so that each PR is more atomic and understandable in its own context.

I agree but I have seen huge PRs in winutil with clear description. If multiple people like you guys go through the code base and approve it won't be a problem for Chris. Let me finish this for now and then decide whether we can split it to smaller ones. The thing is there are minor changes in every file which don't need a separate PR so I thought instead of creating separate PRs for each and every file or change why not create a single PR with all the rust changes included. Still it's not too late, this is WIP and can be split into separate PRs if wanted.

lj3954:
Please drop 6e5fe07, it duplicates #935.

lj3954 (changes requested):
I think this is more concise.

tui/src/filter.rs
```diff
@@ -78,31 +78,29 @@ impl Filter {
...
+            let input = self.search_input.iter().collect::<String>().to_lowercase();
+            self.items.iter().find_map(|item| {
+                let item_name_lower = &item.node.name.to_lowercase();
+                if item_name_lower.starts_with(&input) {
+                    Some(item_name_lower[input.len()..].to_string())
+                } else {
+                    None
+                }
+            })
         }
```
lj3954, line 103:
```suggestion
        });
```

Alongside my other suggested change. A semicolon should be suffixed for good practice since this is a statement, rather than just an expression which happens to return a unit type.

tui/src/filter.rs
```diff
@@ -78,31 +78,29 @@ impl Filter {
...
+        } else {
+            let input = self.search_input.iter().collect::<String>().to_lowercase();
+            self.items.iter().find_map(|item| {
+                let item_name_lower = &item.node.name.to_lowercase();
+                if item_name_lower.starts_with(&input) {
+                    Some(item_name_lower[input.len()..].to_string())
+                } else {
+                    None
+                }
+            })
```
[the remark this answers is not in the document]
jeevithakannan2: @lj3954 Can you open a PR to my branch I think your suggestion here have errors ( might be because you used your browser )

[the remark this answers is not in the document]
jeevithakannan2:
```rust
let render_completion = !(self.items.is_empty() || self.search_input.is_empty());
        self.completion_preview = render_completion.then(|| {
            let input = self.search_input.iter().collect::<String>().to_lowercase();
            self.items
                .iter()
                .find_map(|item| {
                    let item_name_lower = item.node.name.to_lowercase();
                    (item_name_lower.starts_with(&input))
                        .then_some(item_name_lower[input.len()..].to_string())
                })
                .unwrap_or("".to_string())
        });
```

Is this the suggestion?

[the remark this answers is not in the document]
lj3954: Whoops, the suggestion with then() wouldn't work given that we're returning an optional value.

tui/src/filter.rs
```diff
@@ -78,31 +78,29 @@ impl Filter {
...
+            None
+        } else {
+            let input = self.search_input.iter().collect::<String>().to_lowercase();
+            self.items.iter().find_map(|item| {
+                let item_name_lower = &item.node.name.to_lowercase();
+                if item_name_lower.starts_with(&input) {
+                    Some(item_name_lower[input.len()..].to_string())
+                } else {
+                    None
+                }
```
lj3954, line 101:
```suggestion
                let item_name_lower = item.node.name.to_lowercase();
                (item_name_lower.starts_with(&input))
                    .then_some(item_name_lower[input.len()..].to_string())
```

lj3954 (changes requested):

tui/src/filter.rs
jeevithakannan2: Yupp !!!

jeevithakannan2:
@ChrisTitusTech A new crates.io publish should be made for the CI/CD pipeline and for the program to work due to the merge of https://github.com/ChrisTitusTech/linutil/pull/949

lj3954:
Making the theme functions const doesn't accomplish anything useful; they'll still be evaluated at runtime since the theme is not known at compile time.

lj3954:
We also should not be merging PRs like this. I find it unlikely Chris will accept anything with a 1500 line diff.

jeevithakannan2:
> Making the theme functions const doesn't accomplish anything useful; they'll still be evaluated at runtime since the theme is not known at compile time.

It may not be now. But if we decide to change how themes work instead of passing the default theme at run time making it as default during compile time will make this more efficient.

jeevithakannan2:
> We also should not be merging PRs like this. I find it unlikely Chris will accept anything with a 1500 line diff.

I'm aware of that. What can we do? Do I create 9 separate PRs for each commit?

cartercanedy:
> > Making the theme functions const doesn't accomplish anything useful; they'll still be evaluated at runtime since the theme is not known at compile time.
> 
> It may not be now. But if we decide to change how themes work instead of passing the default theme at run time making it as default during compile time will make this more efficient.

I don't mind the idea of marking methods const when compatible, but I don't see any way that the theme will ever become a compile-time constant as long as it remains configurable at runtime.

cartercanedy:
> > We also should not be merging PRs like this. I find it unlikely Chris will accept anything with a 1500 line diff.
> 
> I'm aware of that. What can we do? Do I create 9 separate PRs for each commit?

Yes

adamperkowski:
man that diff is HUGE
@jeevithakannan2, the conflict-creator

jeevithakannan2:
> man that diff is HUGE
> @jeevithakannan2, the conflict-creator

I know we have put a lot of time into this. Separating the PRs will do the same. But this makes the program more efficient in many areas. Can't complain.

adamperkowski:
Didn't expect you to actually merge this huge diff but ok.

PR #938, opened 2024-11-12
make mouse interaction wrap back

nnyyxxxx:
## Type of Change
- [x] New feature

## Description
makes mouse interaction wrap back on last / first entry when scroll is attempted again in that direction, see video for a better explanation

we have this same feature for regular keyboard movement, btw

resolves #939 

## Testing
video is pretty laggy, you are warned !!

https://github.com/user-attachments/assets/2d040821-ba7c-41cf-8a74-e446155d3e6d

PR #942, opened 2024-11-12
Revert #913 "feat: Mouse scroll"

### LINKED ISSUE #936: tui is noticably laggy with mouse interaction
title

### LINKED ISSUE #939: mouse interaction should wrap back on last / first entry just like keyboard interaction
#938

### LINKED ISSUE #941: mouse interaction should be off by default and have a toggle
#937

nnyyxxxx:
Reverts ChrisTitusTech/linutil#913

@ChrisTitusTech this feature lags the tui and increases cpu usage by a crazy amount, i think this should be reverted.
debug: ( much worse )
image
release: ( a little better but still noticeable)
image

@ChrisTitusTech take a look at #936 for more information regarding the bugs that came with the merge of #913 

closes #938
closes #937 
closes #946 
resolves #936 
resolves #941 
resolves #939 

after merging this i think a new prerelease should be made IF this feature was included in the latest one.

Also to bake this up even more, I think tui's shouldn't have mouse interaction in the first place and I think you can agree with me on that one, BECAUSE It's ran in the terminal (1) It's too resource intensive to provide support for it (2), please consider merging this.

## @ChrisTitusTech very important !! check out #946 it provides a major improvement towards this feature that includes more mouse interaction as seen in the video provided.

If you're not willing to merge this, take a look at :
#938
#937
#946

They both provide QOL improvements towards this feature.

If you still want to keep this feature inside of Linutil then this will have to be heavily optimized and I don't think It's possible.

### DIFF
--- a/tui/src/confirmation.rs
+++ b/tui/src/confirmation.rs
@@ -3,7 +3,7 @@ use std::borrow::Cow;
 use crate::{float::FloatContent, hint::Shortcut};
 
 use ratatui::{
-    crossterm::event::{KeyCode, KeyEvent, MouseEvent, MouseEventKind},
+    crossterm::event::{KeyCode, KeyEvent},
     layout::Alignment,
     prelude::*,
     widgets::{Block, Borders, Clear, List},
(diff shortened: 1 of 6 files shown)


nnyyxxxx:
only a 6% loss, and the performance is about the same

jeevithakannan2:
Please check with previous a old release before implementation of mouse scroll. I don't see any difference in CPU usage before and after https://github.com/ChrisTitusTech/linutil/pull/913

nnyyxxxx:
i already have

PR #964, opened 2024-11-22
docs(README): wrong crate name in `cargo install` instructions

### LINKED ISSUE #963: Cargo installation instructions incorrect
## Describe the bug

The cargo installation instructions state to run `cargo install linutil` but that is incorrect, it should be `cargo install linutil_tui`.  (See additional notes below as well.)

## To Reproduce

Look at the crates.io page.  Try copy pasting the cargo install line and run it.

## Expected behavior

Copy pasting the command to install should work

## Checklist
- [X] I checked for duplicate issues.

Additional notes: 

The version currently on crates.io does not compile anyway because someone broke semver.  You actually need to run `cargo install linutil_tui --locked` or it won't compile.

adamperkowski:
Resolves #963

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ sudo zypper install linutil
 Linutil can be installed via [Cargo](https://doc.rust-lang.org/cargo) with:
 
 ```bash
-cargo install linutil
+cargo install linutil_tui
 ```
 
 Note that crates installed using `cargo install` require manual updating with `cargo install --force` (update functionality is [included in LinUtil](https://christitustech.github.io/linutil/userguide/#applications-setup))


jeevithakannan2 (approved):

README.md
```diff
@@ -95,7 +95,7 @@ sudo zypper install linutil
 Linutil can be installed via [Cargo](https://doc.rust-lang.org/cargo) with:
 
 ```bash
-cargo install linutil
+cargo install linutil_tui
```
OvermindDL1, line 98: Should probably add `--locked` to it as well since the version currently on crates cannot compile because one of the dependencies broke semver. Using `--locked` will have the install only use the dependent versions that are in the Cargo.lock file rather than the latest semver matching ones.

(The broken dependency really needs to fix that, `cargo-semver-checks` exists for a reason after all, lol)

OvermindDL1 (changes requested):

README.md
adamperkowski: Thanks for pointing that out. I really have no idea what's going on with versioning here. I'll try to fix all this today but @ChrisTitusTech IMO you really should check if crates 100% work before releasing.

I'm not sure if this will work though, as `Cargo.lock` is not included in the packages.
https://github.com/ChrisTitusTech/linutil/blob/ab7a67087d8387d38cbc839c4725aaf89a70675a/tui/Cargo.toml#L10
https://github.com/ChrisTitusTech/linutil/blob/ab7a67087d8387d38cbc839c4725aaf89a70675a/core/Cargo.toml#L8

adamperkowski:
@ChrisTitusTech Why is the version in `Cargo.toml` = `24.9.28` on the `24.10.31` branch?
https://github.com/ChrisTitusTech/linutil/blob/00914ab05dd64b893e05d1e9b43d8b9e84a7918c/Cargo.toml#L3

lj3954:
> @ChrisTitusTech Why is the version in `Cargo.toml` == `24.9.28` on the `24.10.31` tag?
> 
> https://github.com/ChrisTitusTech/linutil/blob/00914ab05dd64b893e05d1e9b43d8b9e84a7918c/Cargo.toml#L3

#965 will resolve this issue.

ChrisTitusTech (approved):

PR #995, opened 2025-01-11
chore: remove redundant files

adamperkowski:
ref https://github.com/ChrisTitusTech/linutil/security/dependabot


adamperkowski:
i'm also wondering if we really need [those two](https://github.com/ChrisTitusTech/linutil/tree/main/docs) now with other repos for docs

Issue #996, opened 2025-01-11
Issues with `core/tabs/utils/ssh.sh`


koibtw:
First of all, shellcheck.
Second of all, those quotes actually unquote the variable:
https://github.com/ChrisTitusTech/linutil/blob/d91a44efabe09394a60bdbd84099e48d57726720/core/tabs/utils/ssh.sh#L65-L69
Third of all, `$ESCALATION_TOOL` in this case is being expanded on the client side...
The list goes on. We need someone to refactor that whole script.

technicks89:
@koibtw, this looks looks like it was fixed in pull request [1233](https://github.com/ChrisTitusTech/linutil/pull/1233). Can we close this issue

FluffyPunk:
As I see from #1233, initial issue is fixed, so /close

PR #997, opened 2025-01-11
refact: fix (almost) all shellcheck errors

adamperkowski:
the only files i didn't fix:
- core/tabs/utils/ssh.sh (ref #996)
- core/tabs/system-setup/arch/server-setup.sh


adamperkowski:
honestly i think we should get rid of shfmt

ChrisTitusTech (approved):

ChrisTitusTech:
Agree we should probably dump shfmt.

PR #1006, opened 2025-01-20
Change libvirtd service to start instead of enabling

jeevithakannan2:
## Type of Change
- [x] Refactoring

## Description
- A 5-second delay in shutdown or reboot is caused by a libvirt issue when libvirtd service is running.
- Refactored to start the service without enabling it by default, giving users the flexibility to decide whether to enable it.
https://bbs.archlinux.org/viewtopic.php?id=301825

## Testing
- Tested on Arch Linux

## Checklist

### DIFF
--- a/core/tabs/system-setup/arch/virtualization.sh
+++ b/core/tabs/system-setup/arch/virtualization.sh
@@ -60,7 +60,7 @@ setupLibvirt() {
         fi
     done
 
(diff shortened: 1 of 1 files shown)


ChrisTitusTech:
enable --now enables it on startup and starts it. 
This would break the autostart.

PR #961, opened 2024-11-21
fix: scroll down not reaching the bottom entry

jeevithakannan2:
## Type of Change
- [x] Bug fix

## Description
- Fix scroll down not reaching bottom entry in a sub category like communication apps, developer tools, etc.

## Testing
- Tested on Arch Linux without issues.

## Issues
- Resolves #1009 
- Closes #1010 

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #935, opened 2024-11-11
Fixes #934

zdiff:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [ ] UI/UX improvement

## Description
Fixes `linutil --help` command

### Before PR
```shell
./target/release/linutil --help
```
```shell
error: unexpected argument found
```
### After PR
```shell
 ./target/release/linutil --help
```
```shell
Usage: 

Options:
  -c, --config <CONFIG>      Path to the configuration file
  -t, --theme <THEME>        Set the theme to use in the application [default: default] [possible values: default, compatible]
  -y, --skip-confirmation    Skip confirmation prompt before executing commands
      --override-validation  Show all available options, disregarding compatibility checks (UNSAFE)
      --size-bypass          Bypass the terminal size limit
  -h, --help                 Print help
```

## Issues / other PRs related
- Resolves #934

## Checklist
- [x] My code adheres to the coding and style guidelines of the project.
- [x] I have performed a self-review of my own code.
- [ ] I have commented my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the documentation.
- [x] My changes generate no errors/warnings/merge conflicts.

PR #1015, opened 2025-02-02
refactor: misc restructuring

nnyyxxxx:
Features:

Separate cli flags and put them in 
  their own module.

Add a rustfmt configuration file for 
  grouping imports into crates (so we 
  dont have to do this manually, and it 
  seems we were already doing it manually 
  so might as well)

Use comments to describe cli flags 
  instead of using ugly syntax (they both 
  work the same but one is less ugly and 
  more readable)

Add a --mouse flag to enable mouse 
  interaction (mouse interaction is now 
  off by default)

Add a --bypass-root flag to disable 
  the annoying popup when you run the 
  utility as root

Fix an issue where you can't reach the 
  bottom of the list in a subdir (i also 
  reduced the nesting in those functions 
  as well for readability)

Add help feature to clap dependency to 
  enable --help / -h

Add usage feature to clap dependency
  to enable usage example when running
  with --help / -h

Remove CLI arg examples from readme, 
  and add instructions on how to view 
  them on CLI to prevent it from bloating 
  up the readme

Explanation 1:

Refactoring is necessary for reducing 
maintenance burden, improving code 
readability and quality, and fixing bugs. 
Please do not outright deny this pull 
request because it has "refactor" in the 
title; these changes are needed.

Explanation 2:

Some changes in this pull request 
  duplicate the fixes from #961 
  while the changes are not 
  inherently duplicated, I 
  had refactored the 2 func's 
  to improve readability and 
  ended up fixing the same bug. 
  I had commented under that PR 
  asking if I should drop the changes 
  but I have not received a response yet 
  (at the time of writing)


nnyyxxxx:
I also think we should go through and reduce nesting across the codebase as well, the code isn't very readable right now

nnyyxxxx:
@lj3954 @JustLinuxUser @cartercanedy @adamperkowski feel free to review

tui/src/root.rs
```diff
@@ -8,10 +8,14 @@ This means you have full system access and commands can potentially damage your
 Please proceed with caution and make sure you understand what each script does before executing it.";
 
 #[cfg(unix)]
-pub fn check_root_status<'a>() -> Option<FloatingText<'a>> {
-    (Uid::effective().is_root()).then_some(FloatingText::new(
-        ROOT_WARNING.into(),
-        "Root User Warning",
-        true,
-    ))
+pub fn check_root_status<'a>(bypass_root: bool) -> Option<FloatingText<'a>> {
```
cartercanedy, line 11:
```suggestion
pub fn check_root_status(bypass_root: bool) -> Option<FloatingText<'static>> {
```

nnyyxxxx: thx!

JustLinuxUser:
I still haven't reviewed the code, but the --help problem is because someone disabled default-features in clap in Cargo.toml, so a regression introduced in #922 . Just add `help` in the feature list, you can learn more [here](https://docs.rs/clap/latest/clap/_features/index.html)

nnyyxxxx:
> I still haven't reviewed the code, but the --help problem is because someone disabled default-features in clap in Cargo.toml, so a regression introduced in #922 . Just add `help` in the feature list, you can learn more [here](https://docs.rs/clap/latest/clap/_features/index.html)

thanks

tui/src/state.rs
```diff
@@ -598,23 +609,47 @@ impl AppState {
...
-            if selected == self.filter.item_list().len() - 1 {
-                self.selection.select_first();
-            } else {
-                self.selection.select_next();
-            }
-        }
+        let Some(selected) = self.selection.selected() else {
+            return;
+        };
+        let list_len = if !self.at_root() {
```
JustLinuxUser, line 615: `let list_len = self.filter.item_list().len() + !self.at_root() as usize` works too, same in scroll up

tui/src/state.rs
```diff
@@ -598,23 +609,47 @@ impl AppState {
     }
 
     fn scroll_down(&mut self) {
-        if let Some(selected) = self.selection.selected() {
-            if selected == self.filter.item_list().len() - 1 {
```
JustLinuxUser, line 602: Honestly, I think just replacing `self.filter.item_list().len() - 1` with the list_len from your commit would look cleaner and more consistent with the rest of the code

JustLinuxUser (changes requested):
I like the changes, but some small tweaks would improve them even more, especially the scroll up, and down thing, I would prefer something like:
```rust
    fn scroll_down(&mut self) {
        let list_len = self.filter.item_list().len() + !self.at_root() as usize;
        if let Some(selected) = self.selection.selected() {
            if selected == list_len - 1 {
                self.selection.select_first();
            } else {
                self.selection.select_next();
            }
        }
    }

    fn scroll_up(&mut self) {
        if let Some(selected) = self.selection.selected() {
            if selected == 0 {
                self.selection.select_last();
            } else {
                self.selection.select_previous();
            }
        }
    }

```

JustLinuxUser:
Btw, the --help thing, you might want to coordinate that with @zdiff, the creator of #935 pr, and the creator of the PR that created the regression in the first place

nnyyxxxx:
> I like the changes, but some small tweaks would improve them even more, especially the scroll up, and down thing, I would prefer something like:
> 
> ```rust
>     fn scroll_down(&mut self) {
>         let list_len = self.filter.item_list().len() + !self.at_root() as usize;
>         if let Some(selected) = self.selection.selected() {
>             if selected == list_len - 1 {
>                 self.selection.select_first();
>             } else {
>                 self.selection.select_next();
>             }
>         }
>     }
> 
>     fn scroll_up(&mut self) {
>         if let Some(selected) = self.selection.selected() {
>             if selected == 0 {
>                 self.selection.select_last();
>             } else {
>                 self.selection.select_previous();
>             }
>         }
>     }
> ```

to be honest, i like my version better. Maybe its based on style preferences, but i prefer flat, easy to follow, code that has clean separation.

nnyyxxxx:
> Btw, the --help thing, you might want to coordinate that with @zdiff, the creator of #935 pr, and the creator of the PR that created the regression in the first place

I fixed the issue in d1444bd094fa5d6ed43a12b1933fbed49e4f07ae 5164f1493b7d11b09431e9ead51f67a0f9e60d2e

Regardless he should see this pull request since you mentioned him

tui/src/state.rs
nnyyxxxx: I'm going to resolve these conversations since these changes might get dropped depending on the answer in #961

lj3954:
> I like the changes, but some small tweaks would improve them even more, especially the scroll up, and down thing, I would prefer something like:
> 
> ```rust
>     fn scroll_down(&mut self) {
>         let list_len = self.filter.item_list().len() + !self.at_root() as usize;
>         if let Some(selected) = self.selection.selected() {
>             if selected == list_len - 1 {
>                 self.selection.select_first();
>             } else {
>                 self.selection.select_next();
>             }
>         }
>     }
> 
>     fn scroll_up(&mut self) {
>         if let Some(selected) = self.selection.selected() {
>             if selected == 0 {
>                 self.selection.select_last();
>             } else {
>                 self.selection.select_previous();
>             }
>         }
>     }
> ```

I dislike the boolean to integer cast. It makes it more difficult to reason about the code.

JustLinuxUser:
> > I like the changes, but some small tweaks would improve them even more, especially the scroll up, and down thing, I would prefer something like:
> > ```rust
> >     fn scroll_down(&mut self) {
> >         let list_len = self.filter.item_list().len() + !self.at_root() as usize;
> >         if let Some(selected) = self.selection.selected() {
> >             if selected == list_len - 1 {
> >                 self.selection.select_first();
> >             } else {
> >                 self.selection.select_next();
> >             }
> >         }
> >     }
> > 
> >     fn scroll_up(&mut self) {
> >         if let Some(selected) = self.selection.selected() {
> >             if selected == 0 {
> >                 self.selection.select_last();
> >             } else {
> >                 self.selection.select_previous();
> >             }
> >         }
> >     }
> > ```
> 
> I dislike the boolean to integer cast. It makes it more difficult to reason about the code.

It's ok, if you want, you can still use the if, so it would become:
```rust
        let list_len = self.filter.item_list().len() + if self.at_root() { 0 } else { 1 };
```
or even
```rust
let list_len = if self.at_root {self.filter.item_list().len()} else {self.filter.item_list().len() + 1}
```
like you had before, choose whatever you like best
```

File: tui/src/state.rs [hunk 4]
```diff
@@ -598,23 +609,47 @@ impl AppState {
...
+            return;
+        };
+
+        let next_selection = if selected == 0 {
+            list_len - 1
+        } else {
+            selected - 1
+        };
+        self.selection.select(Some(next_selection));
     }
```
adamperkowski, line 653: the shared functionality in those two could be merged into one function

adamperkowski (changes requested):

nnyyxxxx:
@JustLinuxUser take a look at [f28e085](https://github.com/ChrisTitusTech/linutil/pull/1015/commits/f28e0855b6c0e4747637cb5706e3154a5113ed51) and see if you like it better

adamperkowski (approved):
the scope shouldn't be explicitly tui

JustLinuxUser:
> @JustLinuxUser take a look at [f28e085](https://github.com/ChrisTitusTech/linutil/pull/1015/commits/f28e0855b6c0e4747637cb5706e3154a5113ed51) and see if you like it better

That does look good, ty

nnyyxxxx:
oops

nnyyxxxx:
wrong repo :sob:

ChrisTitusTech (approved):

PR #1017, opened 2025-02-02
perf: increase rendering efficiency

JustLinuxUser:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Hotfix
- [ ] Security patch
- [x] UI/UX improvement

## Description
This small PR, makes, so that the UI is not re-rendered, unless an event is detected. The event might be a terminal event (keypress, mouse moved, resize, etc.)
Or if a terminal is open, if some new input came in from the terminal

## Testing
Before: 50% CPU usage on my machine with debug on
After: 1% CPU usage on my machine with debug on

## Impact
It doesn't impact the performance of the app, it will run just as fast as before, but it affects it's efficiency, so it's not re-rendered unless it should be.

## Additional Information
I am not aware of any possible side effects this might create, but if in the future we want to implement something like an animation system, it will complicate it a little.

### DIFF
--- a/tui/src/main.rs
+++ b/tui/src/main.rs
@@ -23,10 +23,12 @@ use ratatui::{
     },
     Terminal,
 };
+use running_command::TERMINAL_UPDATED;
 use state::AppState;
 use std::{
     io::{stdout, Result, Stdout},
     path::PathBuf,
+    sync::atomic::Ordering,
     time::Duration,
 };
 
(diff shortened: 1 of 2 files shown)


adamperkowski (approved):
not sure i'm 100% with you on using atomic but still the improvement is huge. thanks!

JustLinuxUser:
> not sure i'm 100% with you on using atomic but still the improvement is huge. thanks!

I used an atomic, because: it's accessed from a different thread, 2 it's basically free on x86, 3, if it's a global, we can later rename it to SHOULD_REDRAW, and use it when ever needed.

But I would like to know why you don't like the usage of the global atomic. Mb there is a better solution that I am not seeing?. Ty

adamperkowski:
> But I would like to know why you don't like the usage of the global atomic. Mb there is a better solution that I am not seeing?. Ty

not saying i don't like it being there. it's more about when i see atomic i'm like:
- cache issues
- memory ordering
- synchronization costs
- damn deadlocks again
- we have alternatives, higher level abstractions in rust

\+ it might be a little bit overkill
but still, the effects are very much positive and that's what matters :)

ChrisTitusTech (approved):

PR #1036, opened 2025-02-23
fix: reflector returning mirrors that are down

### LINKED ISSUE #1035: [BUG] If a critical mirror is down, the whole Arch Server Setup fails
## Describe the bug
If a critical mirror is down, the whole Arch Server Setup fails. Looking into the mirrors that are set up by the script it only picks the closest servers.

## To Reproduce
1. Choose the Arch Server Setup option in linutil
2. Follow the process answering all the prompts
3. After setting the closest mirrors, the script tries to download packages
4. If there happens to be a mirror that is down (Error 404), the install fails

## Expected behavior
Arch Server should be installed without issue.
Maybe add additional mirrors that are known to be reliable even if they may not be as fast.

## Log output
```
base-devel-1-2-any downloading...
error: failed retrieving file 'linux-lts-6.12.15-1-x86_64.pkg.tar.zst.sig' from za.mirrors.cicku.me : The requested URL returned error: 404
warning: failed to retrieve some files
error: failed to commit transaction (unexpected error)
Errors occurred, no packages were upgraded.
==> ERROR: Failed to install packages to new root

Generated /etc/fstab:

# /dev/vda3
UUID=338ab620-b83d-43be-a6f0-cd376c26acfd       /               ext4            rw,relatime     0 1

# /dev/vda2 LABEL=EFIBOOT
UUID=EEA6-350A          /boot           vfat            rw,relatime,fmask=0022,dmask=0022,codepage=437,iocharset=ascii,shortname=mixed,utf8,errors=remount-ro   0 2


-------------------------------------------------------------------------
GRUB BIOS Bootloader Install & Check
-------------------------------------------------------------------------

-------------------------------------------------------------------------
Checking for low memory systems <8G
-------------------------------------------------------------------------
chroot: failed to run command ‘/bin/bash’: No such file or directory
```

## Additional context
Above is a snippet from where one of the mirrors seems to be down and cannot download the kernel package

## Checklist

jeevithakannan2:
## Type of Change
- [x] Bug fix

## Description
Add --score 5 to return mirrors that have score less than 5 making sure that most of the returned mirrors are active

## Issues / other PRs related
- Resolves #1035 

## Checklist

### DIFF
--- a/core/tabs/system-setup/arch/server-setup.sh
+++ b/core/tabs/system-setup/arch/server-setup.sh
@@ -316,7 +316,11 @@ echo -ne "
                     Setting up $iso mirrors for faster downloads
 -------------------------------------------------------------------------
 "
-reflector -a 48 -c "$iso" -f 5 -l 20 --sort rate --save /etc/pacman.d/mirrorlist
+reflector -a 48 -c "$iso" --score 5 -f 5 -l 20 --sort rate --save /etc/pacman.d/mirrorlist
+if [[ $(grep -c "Server =" /etc/pacman.d/mirrorlist) -lt 5 ]]; then #check if there are less than 5 mirrors
+    cp /etc/pacman.d/mirrorlist.bak /etc/pacman.d/mirrorlist
+fi
+
 if [ ! -d "/mnt" ]; then
     mkdir /mnt
 fi
--- a/core/tabs/system-setup/arch/server-setup.sh
+++ b/core/tabs/system-setup/arch/server-setup.sh
@@ -183,6 +183,7 @@ keymap () {
     echo -ne "
     Please select key board layout from this list"
     # These are default key maps as presented in official arch repo archinstall
+    # shellcheck disable=SC1010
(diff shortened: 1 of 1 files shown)


core/tabs/system-setup/arch/server-setup.sh
```diff
@@ -316,7 +316,11 @@ echo -ne "
                     Setting up $iso mirrors for faster downloads
 -------------------------------------------------------------------------
 "
-reflector -a 48 -c "$iso" -f 5 -l 20 --sort rate --save /etc/pacman.d/mirrorlist
+reflector -a 48 -c "$iso" --score 5 -f 5 -l 20 --sort rate --save /etc/pacman.d/mirrorlist
```
jeevithakannan2, line 319: Instead of `--score 5` we could also use `--completion-percent 100` which makes sure all the packages are available. But score also calculated with (hours delay + average duration + standard deviation) / completion percentage. We can use either one to solve this issue

ChrisTitusTech (approved):

Issue #1025, opened 2025-02-12
bug: TUI panics after tab completion


nnyyxxxx:
TUI panics after tab completing if you press a key

PR #1083, opened 2025-06-28
fix: panic using any key other than enter after tab completion in search

jeevithakannan2:
## Type of Change
- [x] Bug fix

## Description
- Might panic if non ascii characters are used for scipt names (Which we won't be doing for now)

## Issues / other PRs related
- Resolves https://github.com/ChrisTitusTech/linutil/issues/1025

## Checklist

### DIFF
--- a/tui/src/filter.rs
+++ b/tui/src/filter.rs
@@ -90,11 +90,10 @@ impl Filter {
             None
         } else {
             let input = self.search_input.iter().collect::<String>().to_lowercase();
-            self.items.iter().find_map(|item| {
-                let item_name_lower = item.node.name.to_lowercase();
-                (item_name_lower.starts_with(&input))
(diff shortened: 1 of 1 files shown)


jeevithakannan2:
@ChrisTitusTech #1087 fixes the rust checks

lj3954:
This works, but I don't understand why. The completion itself is done in another function, and doesn't even use the completion preview.

lj3954:
Oh, I see. It's because we're using then_some() rather than then(), which evaluates the expression (string slicing) even if the case is false. Your solution changes the behaviour (completion is non-lowercased), while the issue could be fixed instead by just replacing `then_some(` with `then(|| `, while maintaining current behaviour.

The implementation of completion in general seems a bit faulty in the first place, I'll probably submit a PR soon with some other changes to it.

PR #1097, opened 2025-07-21
Improve completions

### LINKED ISSUE #1025: bug: TUI panics after tab completion

lj3954:
…y to search

## Type of Change
- [x] Bug fix
- [x] Refactoring

## Description
Replace completion_preview in the search implementation with completion, directly appending the completion to the search when accepted rather than searching again for a matching item and replacing the search input with that item. This improves simplicity by keeping all logic for finding the completion in one function, and makes the completion behave as a user would expect, simply appending its exact text to the search rather than replacing all characters, modifying cases in the process. Improve logic for finding said completions by using char Iterators rather than starts_with() and string slicing, fixing a preexisting issue (#1025) caused by eager evaluation of string slicing without prior length checks, as well as any potential issue with multibyte characters in the search field.

## Testing
Tab correctly appends any completion to the search input, and no panics occur when pressing other keys.

## Issues / other PRs related

- Resolves #1025 
- Closes #1083

## Checklist

### DIFF
--- a/tui/src/filter.rs
+++ b/tui/src/filter.rs
@@ -20,8 +20,8 @@ pub struct Filter {
     in_search_mode: bool,
     input_position: usize,
     items: Vec<ListEntry>,
-    // No complex string manipulation is done with completion_preview so we can use String unlike search_input
-    completion_preview: Option<String>,
+    // No complex string manipulation is done with completion so we can use String unlike search_input
+    completion: Option<String>,
 }
 
 impl Filter {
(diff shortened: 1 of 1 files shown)


lj3954:
Linting issues should all be unrelated to this PR, clippy just had some reclassifications in the last few rust versions.

ChrisTitusTech (approved):

PR #1108, opened 2025-08-09
fix(ci): various security vulnerabilities

adamperkowski:
- fix potential code injection in `issue-slash-commands.yaml` by using
  single quotes insted of double, see
  https://www.gnu.org/software/bash/manual/html_node/Double-Quotes.html
- declare permissions for all workflows, see
  https://codeql.github.com/codeql-query-help/actions/actions-missing-workflow-permissions


ChrisTitusTech (commented):
looks good

PR #1143, opened 2025-10-08
feat(ci): automatic PR labeling

koibtw:
so i decided to make labeling prs a bit easier and this workflow automatically syncs issue labels (excluding status labels so those not containing `:`) to a newly created pr (if it has a closing keyword for that issue of course).
this WILL NOT always be accurate so we still need to review manually.
it's not a replacement, just makes stuff easier.

[fully tested on my fork](https://github.com/adamperkowski/linutil-dev/pull/16)

> [!IMPORTANT]
>

### DIFF
--- a/.github/workflows/pr-labels.yaml
+++ b/.github/workflows/pr-labels.yaml
@@ -1,0 +1,54 @@
+name: Label PR
+
+on:
+  pull_request:
+    types: [opened]
+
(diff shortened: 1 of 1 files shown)


koibtw:
id like to hear ur thoughts @ChrisTitusTech @afonsofrancof

PR #1242, opened 2026-04-01
fix: harden SSH and Samba setup flows

Abs313a:
## Type of Change
- [ ] New feature
- [X] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [X] UI/UX improvement

## Summary

- Harden the SSH Samba and SSH Commands utilities so setup and teardown behave reliably across normal user flows and distro differences.

## Issues addressed
1. SSH Samba accepted arbitrary Samba usernames even though `smbpasswd -a` only works for an existing local Unix account, which caused setup failure for valid-looking but nonexistent usernames.
2. SSH Samba asked for a Samba password twice, reported success, then ignored those values and invoked interactive `smbpasswd -a` anyway.
3. SSH Samba did not validate smb.conf with `testparm` before enabling services.
4. SSH Samba handled existing configs poorly by not recovering the configured share path from `smb.conf` for its success output.
5. Firewall setup relied on UFW profile names like OpenSSH and Samba, which are not portable across systems.
6. SSH Commands assumed remote privilege escalation via `sudo -S`, even though the shared escalation path also allows `doas`.
7. SSH Commands edited `~/.ssh/config` using fixed line ranges, which was inconsistent with the host blocks it actually writes.
8. SSH Samba had no built-in way to safely remove previously applied service, firewall, or local SSH setup.

### DIFF
--- a/core/tabs/utils/samba-ssh-setup.sh
+++ b/core/tabs/utils/samba-ssh-setup.sh
@@ -1,12 +1,17 @@
 #!/bin/sh -e
 
 # Load common script functions
-. ../common-script.sh  
+# shellcheck disable=SC1091
+# shellcheck source=../common-script.sh
+. ../common-script.sh
+# shellcheck disable=SC1091
+# shellcheck source=../common-service-script.sh
 . ../common-service-script.sh
 
 # Function to install packages based on the package manager
 install_package() {
     PACKAGE=$1
+    PACKAGER=${PACKAGER:-}
     if ! command_exists "$PACKAGE"; then
         case "$PACKAGER" in
             pacman)
--- a/core/tabs/utils/samba-ssh-setup.sh
+++ b/core/tabs/utils/samba-ssh-setup.sh
@@ -27,6 +27,161 @@ install_package() {
     fi
 }       
 
+validate_samba_config() {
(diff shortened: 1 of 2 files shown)


Abs313a:
Implemented and validated the SSH/Samba hardening changes.

What changed:
- Hardened `SSH Samba` setup by validating local Samba users before calling `smbpasswd -a`.
- Removed the misleading password-confirmation flow and now rely on the real Samba password step.
- Added `testparm` validation before enabling Samba services.
- Fixed existing-config handling so the configured share path is recovered and shown correctly.
- Added UFW fallback behavior for systems where `OpenSSH` / `Samba` profiles are not available.
- Hardened `SSH Commands` remote auth toggles so they do not assume `sudo -S` on the remote side.
- Fixed local `~/.ssh/config` host entry handling to operate on full host blocks instead of fixed line ranges.
- Added a safe `Remove Setup` flow for SSH/Samba teardown, including SSH service setup, Samba service setup, SSH firewall rules, local SSH host entries, and remove-all.

Validation performed:
- `Setup All`.
- Existing Samba config flow with both `Y` and `N`.
- `Remove all`.
- Full remove -> setup cycle.
- `shellcheck`.
- `cargo clippy --workspace --all-targets --all-features`.
- `typos` (only existing false positives outside these changes).

PR #1250, opened 2026-04-10
pr: Made $TIMEZONE offline fallback "tzselect" which is an included script with "tzdata" https://archlinux.org/packages/core/x86_64/tzdata/

ghost:
## Type of Change
- [ ] New feature
- [ ] Bug fix
- [ ] Documentation update
- [x] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Issues / other PRs related

- Resolves #

### DIFF
--- a/core/tabs/system-setup/arch/server-setup.sh
+++ b/core/tabs/system-setup/arch/server-setup.sh
@@ -161,23 +161,14 @@ timezone () {
     while true; do
         read -r -p "Detected timezone: '$time_zone'. Is this correct? (Y/n) "
         case "${REPLY:-Y}" in
-            [Yy]* )
-                TIMEZONE="${time_zone}"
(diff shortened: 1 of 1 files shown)


hyliancode:
The command `tzselect` is a part of `tzdata`. It is a bit verbose with user input, but it guarantees a successful and accurate timezone. Also this is my first pr, apologies if its done incorrectly.

- `tzdata` is basically on all systems, its required by `glibc` and `python`, so everyone should have `tzselect`, and Arch live ISOs **do** include it
- I prefer to use the `read` commands built-in "REPLY" variable instead creating an unnecessary one
- I was going to remove bash-isms but they are already used other places in the script. Not sure if you were still trying to be POSIX with this script

ChrisTitusTech (approved):

PR #1283, opened 2026-05-15
Add 5 Arch Linux system setup scripts

thepinak503:
## Type of Change
- [X] New feature

## Description

Adds 5 new standalone Arch Linux setup scripts under `core/tabs/system-setup/arch/`:

1. **Pacman Config** - Enables Color, ILoveCandy, ParallelDownloads, VerbosePkgLists, multilib, sets MAKEFLAGS in makepkg.conf
2. **PipeWire Audio** - Installs PipeWire + WirePlumber + pulse/alsa/jack compat + 32-bit libraries
3. **Zram Swap** - Configures compressed RAM swap via zram-generator with zstd compression, sets swappiness=10
4. **Snapper Snapshots** - Snapper + snap-pac + grub-btrfs with hourly Btrfs snapshots (Btrfs-only precondition)
5. **System Maintenance** - Enables paccache.timer, removes orphan packages, cleans system journal

All scripts follow existing conventions (common-script.sh, color vars, checkEnv/checkEscalationTool).

## Issues / other PRs related
Covers community-requested features from discussions (#237, #676) and fills gaps in Arch coverage.

## Screenshots (if applicable)
N/A

### DIFF
--- a/core/tabs/system-setup/arch/snapshot-setup.sh
+++ b/core/tabs/system-setup/arch/snapshot-setup.sh
@@ -17,10 +17,24 @@ checkBtrfs() {
 setupSnapper() {
     "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm snapper snap-pac grub-btrfs
 
-    if [ -d "/.snapshots" ] && mountpoint -q "/.snapshots"; then
-        "$ESCALATION_TOOL" umount /.snapshots 2>/dev/null || true
+    if [ -f "/etc/snapper/configs/root" ]; then
+        printf "%s\n" "${RED}Error: Existing Snapper 'root' configuration detected. Aborting.${RC}" >&2
+        exit 1
     fi
-    "$ESCALATION_TOOL" rm -rf /.snapshots 2>/dev/null || true
+
+    if mountpoint -q "/.snapshots"; then
+        printf "%s\n" "${RED}Error: /.snapshots is actively mounted. Aborting.${RC}" >&2
+        exit 1
+    fi
+
+    if [ -d "/.snapshots" ]; then
+        if "$ESCALATION_TOOL" rmdir "/.snapshots" 2>/dev/null; then
+            printf "%s\n" "${GREEN}Removed empty /.snapshots directory.${RC}"
+        else
+            printf "%s\n" "${RED}Error: /.snapshots exists and is not empty or could not be removed. Aborting to avoid data loss.${RC}" >&2
+            exit 1
+        fi
+    fi
 
     "$ESCALATION_TOOL" snapper -c root create-config /
 
@@ -33,6 +47,5 @@ setupSnapper() {
 }
 
 checkEnv
-checkEscalationTool
 checkBtrfs
 setupSnapper
--- a/core/tabs/system-setup/arch/snapshot-setup.sh
+++ b/core/tabs/system-setup/arch/snapshot-setup.sh
@@ -1,0 +1,38 @@
+#!/bin/sh -e
+
+. ../../common-script.sh
+
+checkBtrfs() {
+    if ! command_exists btrfs; then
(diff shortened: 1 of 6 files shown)


Abs313a (changes requested):
### This section of the script poses a severe risk of permanent data loss on systems that already have Snapper configured:
- `‎core/tabs/system-setup/arch/snapshot-setup.sh`  -  line: 23

### Issues:
- Destructive fallback: If `umount` fails, for example because the mount point is busy or locked by a process, the script immediately proceeds to `rm -rf /.snapshots`. If the filesystem is still mounted, this can attempt to delete live snapshots, subvolumes, and history files.
- Silenced failures: `2>/dev/null || true` hides errors and forces the script to continue blindly, even when it may be deleting user data or breaking an existing Snapper layout.
- No safeguards: There is no validation check for an existing Snapper `root` configuration.

See inline comment below for suggested fix.
Happy to discuss :)

core/tabs/system-setup/arch/snapshot-setup.sh
```diff
@@ -0,0 +1,38 @@
...
+    fi
+}
+
+setupSnapper() {
+    "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm snapper snap-pac grub-btrfs
+
+    if [ -d "/.snapshots" ] && mountpoint -q "/.snapshots"; then
+        "$ESCALATION_TOOL" umount /.snapshots 2>/dev/null || true
+    fi
+    "$ESCALATION_TOOL" rm -rf /.snapshots 2>/dev/null || true
```
Abs313a: #### This prevents accidental data loss by aborting when an existing Snapper root config is present, refusing to modify an actively mounted `/.snapshots`, and only removing the directory with `rmdir` when it is empty. Non-empty snapshot directories now stop the script instead of being recursively deleted.

```suggestion
if [ -f "/etc/snapper/configs/root" ]; then
    printf "%s\n" "${RED}Error: Existing Snapper 'root' configuration detected. Aborting.${RC}" >&2
    exit 1
fi

if mountpoint -q "/.snapshots"; then
    printf "%s\n" "${RED}Error: /.snapshots is actively mounted. Aborting.${RC}" >&2
    exit 1
fi

if [ -d "/.snapshots" ]; then
    if "$ESCALATION_TOOL" rmdir "/.snapshots" 2>/dev/null; then
        printf "%s\n" "${GREEN}Removed empty /.snapshots directory.${RC}"
    else
        printf "%s\n" "${RED}Error: /.snapshots exists and is not empty or could not be removed. Aborting to avoid data loss.${RC}" >&2
        exit 1
    fi
fi
```

core/tabs/system-setup/arch/zram-setup.sh
```diff
@@ -0,0 +1,31 @@
...
+vm.swappiness = 10
+vm.vfs_cache_pressure = 50
+EOF
+
+    printf "%b\n" "${GREEN}zram configured: zstd compression, swappiness=10.${RC}"
+    printf "%b\n" "${YELLOW}Reboot or run: sudo systemctl start systemd-zram-setup@zram0${RC}"
+}
+
+checkEnv
+checkEscalationTool
```
technicks89: There is no need for checkEscalationTool as its run in checkEnv.  Only need to have checkEnv for all scripts

thepinak503:
I've addressed all the feedback:

- Removed all `local` keywords (bashism) — scripts are now pure POSIX `sh`
- Removed redundant `checkEscalationTool` calls (already in `checkEnv`)
- Fixed unquoted `$orphans` — now uses `xargs` piping
- Applied the snapper safety fix: checks for existing config, refuses to modify mounted /.snapshots, only `rmdir` if empty

@Abs313a @technicks89 could you re-review when you get a chance?

technicks89:
This work is now duplicated in #1294.  Request this PR to be closed and make suggested modifications to PR #1294

Issue #1129, opened 2025-09-19
[Feature Request]: Arch Server Install nouveau option


Mav3r1ckHL:
### Is your feature request related to a problem? Please describe.

When I attempted to install the Arch Server setup on a very old Asus N61Vn-A2 which has those weird NVIDIA Optimus configurations (GT 220M), the system fails to cleanly load Arch post installation due to the newer NVIDIA drivers that are being installed as part of the script.  What happens is it would get stuck when it reaches the the Graphical Interface target and hangs indefinitely.

Usually for these older GPU the only option is either nouveau or scouring the AUR for the specific legacy driver (in this case it would be the NVIDIA 340 drivers). 

### Describe the solution you'd like

For the sake of simplicity having another prompt to ask if you have an older NVIDIA GPU and asking the end user if they want to use the nouveau drivers instead of the nvidia-lts drivers during the initial prompts would be ideal.

### Describe alternatives you've considered

An extra possible option is asking which legacy NVIDIA driver you want to install from the AUR with a warning that it is at the user's own risk for instability.  Although most likely from a security point of view I don't see it as a viable option but it is nice to ask if possible.

### Checklist

- [x] I checked for duplicate issues.
- [x] I checked already existing discussions.
- [x] This feature is not included in the roadmap.

koibtw:
> For the sake of simplicity having another prompt to ask if you have an older NVIDIA GPU and asking the end user if they want to use the nouveau drivers instead of the nvidia-lts drivers during the initial prompts would be ideal.

i agree

technicks89:
@FluffyPunk @koibtw, this was merged in and should be closed

PR #1293, opened 2026-05-24
feat: add nouveau option for old NVIDIA GPUs in Arch Server install

thepinak503:
Implements #1129

When an NVIDIA GPU is detected during Arch Server install, a prompt now asks the user to choose between:
- `nvidia-lts` (proprietary, default - same as before)
- `nouveau` (open-source, skips nvidia-lts - for old GPUs that hang with proprietary driver)

Per Arch Wiki recommendation, `xf86-video-nouveau` is not installed - kernel's built-in nouveau + modesetting is preferred.

No risk: default behavior is identical, prompt only appears on NVIDIA GPUs.

### DIFF
--- a/core/tabs/system-setup/arch/server-setup.sh
+++ b/core/tabs/system-setup/arch/server-setup.sh
@@ -536,6 +536,21 @@ fi
 
 gpu_type=$(lspci | grep -E "VGA|3D|Display")
 
+select_nvidia_driver() {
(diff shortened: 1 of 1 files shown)


ChrisTitusTech (approved):
Rebased onto current main, dropping stale generated documentation changes. Bash syntax, ShellCheck, linutil_core tests, and all GitHub checks pass.

PR #1294, opened 2026-05-24
fix: 3 Arch-related script fixes (Waydroid, MyBash pkg mgr, GRUB auto-refresh)

### LINKED ISSUE #1067: Syntax error while installing Waydroid
## Describe the bug
Syntax error as missing "(" in script to install Waydroid

## To Reproduce
1. Go to Applications Setup
2. Click on Watdroid
3. Does not prompt user for password unless manually pressing "Enter" key

## Expected behavior
Prompt for user password and install Waydroid

## Screenshots

Image

## Checklist

### LINKED ISSUE #1220: [Feature Request]: add support for when adding a grub theme that it auto refreshes grub
### Problem Description

it does not auto refresh by runing grub2-mkconfig -o /boot/grub2/grub.cfg

### Proposed solution

to make the install also run grub2-mkconfig -o /boot/grub2/grub.cfg at the end

### Alternatives

_No response_

### Additional Context

_No response_

### Checklist

- [x] This feature is not included in the roadmap.

### LINKED ISSUE #1255: [Feature Request]: Starship, fzf and zoxide are installed without package manager
### Problem Description

Hello, I'm using CachyOS.

I used the Bash Prompt script for testing, but then I realized that starship, fzf and zoxide are not installed with pacman, so that I cannot manage updates for those utilities with pacman.

Actually, I decided that I wanted to revert the script because I prefer to use fish, but with this kind of install that is more difficult to do. It would be better to install those packages using the package manager.

### Proposed solution

Something similar to what is done in the `installDepend` function in `mybash-setup.sh` for dependencies. That approach could be used for starship, fzf and zoxide too. And it would be nice if the user can decide if they want to fallback to non-package-manager install or not.

### Alternatives

An revert script could be written. This could allow users to clean what was done by `mybash-setup.sh` if they don't like the result

### Additional Context

I'm using CachyOS with XFCE.

### Checklist

- [x] This feature is not included in the roadmap.

thepinak503:
## Summary

Three independent fixes for Arch Linux user scripts.

### #1067 — Waydroid grep syntax
Simplified the kernel detection grep in `waydroid-setup.sh`. Old ERE pattern used `(|...)` alternation with empty match — fragile and confusing. Replaced with simple piped greps: `grep '^linux' | grep -v '-headers' | grep -v '-firmware'`.

### #1255 — MyBash: install starship/fzf/zoxide via package manager
`mybash-setup.sh` now tries the system package manager first for starship, fzf, and zoxide. Falls back to curl/git only if the package manager fails. Added a shared `_install_pkg` helper matching the pattern used by `installDepend`.

### #1220 — GRUB theme auto-refresh
`grub-theme.sh` now runs `grub-mkconfig -o /boot/grub/grub.cfg` (or `grub2-mkconfig` for Fedora/RHEL) after installing the theme, so users don't have to manually refresh.

## Security
- All changes use existing `"$ESCALATION_TOOL"` pattern
- Package manager + fallback preserves original behavior
- No new network calls or unsafe operations

Closes #1067
Closes #1255
Closes #1220


core/tabs/system-setup/arch/zram-setup.sh
technicks89, line 1: zram-generator is available for other distros. Makes sense to add them to this and move this outside the arch folder.  I think this makes more sense in the utilities folder. @Abs313a @koibtw Thoughts?

core/tabs/system-setup/arch/pipewire-setup.sh
technicks89, line 1: pipewire is available on most distros and this should also be updated and moved into system-setup root folder

technicks89 (commented):
A few of the tools here would make sense to have available for all Distros.  Updating to support other distros (apt-get/nala, dnf, zypper, etc). move to system-setup root instead of arch folder

thepinak503:
Addressed review feedback from @technicks89:

- **pipewire-setup.sh**: Added cross-distro package manager support (pacman, apt/nala, dnf, zypper, apk, xbps, eopkg) and moved from `arch/` to `system-setup/` root so it's available to all distros.
- **zram-setup.sh**: Same treatment — cross-distro `case $PACKAGER` block (uses `systemd-zram-generator` on Debian/Ubuntu, `zram-generator` elsewhere) and moved to root.
- **tab_data.toml**: Entries moved from the Arch section to root-level `[[data]]` entries so they show up for all distros.

Kept `pacman-config.sh`, `snapshot-setup.sh`, and `system-maintenance.sh` in `arch/` since they use Arch-specific commands/paths (pacman.conf/makepkg.conf, paccache, snap-pac/grub-btrfs).

core/tabs/applications-setup/mybash-setup.sh
```diff
@@ -54,29 +54,38 @@ installFont() {
     fi
 }
 
+_install_pkg() {
+    case "$PACKAGER" in
+        pacman) "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm "$1" ;;
+        apk)    "$ESCALATION_TOOL" "$PACKAGER" add "$1" ;;
+        xbps-install) "$ESCALATION_TOOL" "$PACKAGER" -Sy "$1" ;;
+        *)      "$ESCALATION_TOOL" "$PACKAGER" install -y "$1" ;;
```
technicks89: Recommend change *) to dnf|apt-get|nala|zypper, and using * for flatpak(if available) or note of unsupported package manager

core/tabs/system-setup/pipewire-setup.sh
```diff
@@ -0,0 +1,48 @@
...
+            "$ESCALATION_TOOL" "$PACKAGER" add pipewire pipewire-pulse wireplumber
+            ;;
+        xbps-install)
+            "$ESCALATION_TOOL" "$PACKAGER" -Sy pipewire pipewire-pulse wireplumber
+            ;;
+        eopkg)
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire wireplumber
+            ;;
+        *)
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire wireplumber
```
technicks89: Recommend change *) to flatpak(if available) or note of unsupported package manager

core/tabs/system-setup/zram-setup.sh
```diff
@@ -0,0 +1,60 @@
...
+            "$ESCALATION_TOOL" "$PACKAGER" add zram-generator
+            ;;
+        xbps-install)
+            "$ESCALATION_TOOL" "$PACKAGER" -Sy zram-generator
+            ;;
+        eopkg)
+            "$ESCALATION_TOOL" "$PACKAGER" install -y zram-generator
+            ;;
+        *)
+            "$ESCALATION_TOOL" "$PACKAGER" install -y zram-generator
```
technicks89: Recommend change *) to flatpak(if available) or note of unsupported package manager

technicks89 (changes requested):

thepinak503:
Addressed review feedback from @technicks89's second review:

- **mybash-setup.sh**: Changed `_install_pkg` `*)` fallback to explicitly list `dnf|apt-get|nala|zypper|eopkg` for `install -y`, with `*)` now returning 1 (triggering the existing curl/git fallback) instead of blindly running an unknown package manager.
- **pipewire-setup.sh & zram-setup.sh**: Replaced the `*)` catch-all that blindly ran `$PACKAGER install -y` with an explicit unsupported message + exit, since all known package managers are already handled in specific branches.

core/tabs/system-setup/pipewire-setup.sh
```diff
@@ -0,0 +1,49 @@
...
+        pacman)
+            "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm pipewire pipewire-pulse pipewire-alsa pipewire-jack wireplumber lib32-pipewire
+            ;;
+        apt-get|nala)
+            "$ESCALATION_TOOL" "$PACKAGER" update
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire pipewire-pulse wireplumber pipewire-audio-client-libraries 2>/dev/null || \
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire pipewire-pulse wireplumber
+            ;;
+        dnf)
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire pipewire-pulse wireplumber pipewire-jack-audio-connection-kit
```
technicks89: pipewire-pulse should be pipewire-pulseaudio

core/tabs/system-setup/pipewire-setup.sh
```diff
@@ -0,0 +1,49 @@
...
+
+installPipewirePkgs() {
+    case "$PACKAGER" in
+        pacman)
+            "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm pipewire pipewire-pulse pipewire-alsa pipewire-jack wireplumber lib32-pipewire
+            ;;
+        apt-get|nala)
+            "$ESCALATION_TOOL" "$PACKAGER" update
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire pipewire-pulse wireplumber pipewire-audio-client-libraries 2>/dev/null || \
+            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire pipewire-pulse wireplumber
```
technicks89: Debian (at least 13) has pipewire-jack as well

technicks89 (changes requested):

thepinak503:
Updated pipewire-setup.sh based on feedback:

- **Debian (apt-get/nala):** Added `pipewire-jack` to the install list. Debian 13 (Trixie) ships i            "$ESCALATION_TOOL" "$PACKAGER" install -y pipewire pipewire-pulse wireplumber pipewire-jack-audio-connection-kitt as a separate package. It makes sense to include it. The `2>/dev/null ||` fallback handles older releases gracefully.
- **Fedora (dnf):** Fixed package name from `pipewire-pulse` to `pipewire-pulseaudio`. It is the correct Fedora package name.

technicks89 (approved):

ChrisTitusTech (approved):

PR #1313, opened 2026-06-10
Cachy os fix

technicks89:
## Type of Change
- [X] Bug fix

## Description
When running from curl, cachyos script would fail due to not properly checking if repos were installed

## Issues / other PRs related

- Resolves #


StKob:
As of 25.06.2026 this PR indeed fixes repo installation. Before it would silently fail due to -e flag for sh and incorrect error handling in repo check.

PR #1318, opened 2026-06-22
Add SSD TRIM (fstrim.timer) setup script

emylfy:
## Type of Change
- [x] New feature

## Description

Adds a new System Setup entry, "Enable SSD TRIM". It turns on systemd's `fstrim.timer` (the recommended weekly TRIM, rather than the continuous `discard` mount option) and runs an initial `fstrim -av` so unused blocks are reclaimed right away instead of waiting for the first scheduled run.

The script only installs `util-linux` when `fstrim` is missing (covers the common package managers), exits early if `fstrim.timer` isn't available on the system, and is safe to re-run. I gave the entry a `command_exists systemctl` precondition so it only shows up on systemd systems.

This is useful mainly on Arch and other minimal/DIY installs, where the timer is not enabled out of the box. On distros that already ship it enabled (Ubuntu, recent Fedora) the script is a no-op beyond reconfirming the timer and running the immediate trim.

Changed files: `fstrim-setup.sh`, the matching `tab_data.toml` entry, and the regenerated walkthrough (`cargo xtask docgen`).

## Testing

Tested on Ubuntu 24.04 (aarch64), two consecutive runs.

First run trims the live filesystems:

```
fstrim is already installed
Enabling weekly fstrim.timer...
Running an initial trim on all supported mounted filesystems...
/boot/efi: 1 GiB (1118154752 bytes) trimmed on /dev/sda1
/: 48.2 GiB (51757043712 bytes) trimmed on /dev/sda2
Periodic SSD TRIM is active (weekly via fstrim.timer).
```

Second run, the root filesystem reports only the few blocks freed since (the FAT EFI partition always reports its full size):

```
fstrim is already installed
Enabling weekly fstrim.timer...
Running an initial trim on all supported mounted filesystems...
/boot/efi: 1 GiB (1118154752 bytes) trimmed on /dev/sda1
/: 904 KiB (925696 bytes) trimmed on /dev/sda2
Periodic SSD TRIM is active (weekly via fstrim.timer).
```

`shellcheck` passes.

## Issues / other PRs related
None.

### DIFF
--- a/core/tabs/system-setup/fstrim-setup.sh
+++ b/core/tabs/system-setup/fstrim-setup.sh
@@ -1,0 +1,52 @@
+#!/bin/sh -e
+
+. ../common-script.sh
+
+installFstrim() {
+    # fstrim ships with util-linux
+    if ! command_exists fstrim; then
+        printf "%b\n" "${YELLOW}Installing fstrim (util-linux)...${RC}"
+        case "$PACKAGER" in
+            pacman)
+                "$ESCALATION_TOOL" "$PACKAGER" -S --needed --noconfirm util-linux
+                ;;
+            apk)
+                "$ESCALATION_TOOL" "$PACKAGER" add util-linux
+                ;;
+            xbps-install)
+                "$ESCALATION_TOOL" "$PACKAGER" -Sy util-linux
+                ;;
+            *)
+                "$ESCALATION_TOOL" "$PACKAGER" install -y util-linux
+                ;;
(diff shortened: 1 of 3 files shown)


ChrisTitusTech (approved):
Rebased onto current main and added the required disk-operation warning and non-queueable metadata. linutil_core tests, shell syntax, ShellCheck, doc generation, and diff checks pass locally.

PR #1357, opened 2026-07-28
apps(browsers): Add Trivalent (chromium fork)

nomoregloom:
## Type of Change
- [x] New feature
- [ ] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description

Adds a Linux based, hardened fork of Chromium, similar to Helium browser but focused more on security around LInux specifically rather than debloating. You can check the upstream repo to get a better idea of what it actually does:

https://github.com/secureblue/Trivalent

It is similar to "Vanadium" (GrapheneOS browser). it's basically it, ported to Linux.

Notes: Debian based distros and OpenSUSE Tumbleweed as of now are unsupported.
Arch package is unofficial

### DIFF
--- a/core/tabs/applications-setup/browsers/trivalent.sh
+++ b/core/tabs/applications-setup/browsers/trivalent.sh
@@ -1,0 +1,120 @@
+#!/bin/sh -e
+
+. ../../common-script.sh
+
+installTrivalent() {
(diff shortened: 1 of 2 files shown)


ChrisTitusTech:
too niche

PR #1365, opened 2026-08-04
fix: keep EQ FluidSynth alive without a server

ChrisTitusTech:
Follow-up to #1364 after live Fedora 44 validation exposed a FluidSynth lifecycle issue.

## What changed

- Keep FluidSynth's local shell loop alive through a private mode-0600 FIFO instead of `--server`.
- Remove the FIFO alongside PID cleanup and reject unsafe runtime paths.
- Add `mkfifo` to installer and catalog prerequisites.

## Root cause

FluidSynth 2.5.4 exits normally when `--no-shell` is used without server mode. Server mode keeps it alive but opens TCP port 9800 on all interfaces. The private FIFO preserves the process without a network listener.

## Validation

- Live Fedora 44 installation against EverQuest Legends in native Lutris.
- Verified the ALSA `EQ-Legends` client and port while running.
- Verified no FluidSynth TCP listener.
- Verified watcher-driven shutdown at 65 seconds removed the process, PID file, FIFO, and launch tokens.
- `sh -n`, ShellCheck, checkbashisms, and shfmt pass for all three scripts.
- `cargo fmt --all --check`, `cargo test --no-fail-fast --package linutil_core`, `cargo clippy -- -Dwarnings`, `cargo xtask docgen`, and `git diff --check` pass.

### DIFF
--- a/core/tabs/gaming/eq-legends-midi-start.sh
+++ b/core/tabs/gaming/eq-legends-midi-start.sh
@@ -13,6 +13,7 @@ fi
 STATE_DIR="$STATE_HOME/linutil"
 PID_FILE="$RUNTIME_DIR/fluidsynth.pid"
 LOCK_FILE="$RUNTIME_DIR/state.lock"
+SHELL_FIFO="$RUNTIME_DIR/fluidsynth.stdin"
 PENDING_DIR="$RUNTIME_DIR/pending"
 LOG_FILE="$STATE_DIR/eq-legends-midi.log"
 STOP_HOOK="$HOOK_DIR/eq-midi-stop.sh"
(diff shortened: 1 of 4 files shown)


ChrisTitusTech:
@codex review

ChrisTitusTech:
Reviewed the CodeRabbit nitpick on `command_exists`: `eq-legends-midi-start.sh` is a standalone extracted Lutris lifecycle hook and intentionally does not source `common-script.sh`. Its hook-local dependency checks consistently use `command -v`; replacing only the `mkfifo` check would call an undefined helper at runtime. No change is needed for that approved nitpick.

PR #1376, opened 2026-08-26
Use tealdeer instead of tldr on apt-based distros

karsterr:
## Type of Change
- [ ] New feature
- [x] Bug fix
- [ ] Documentation update
- [ ] Refactoring
- [ ] Security patch
- [ ] UI/UX improvement

## Description
On Debian-based systems, `compile-setup.sh` fails during dependency installation because the `tldr` package is absent from standard APT repositories. Debian package index provides `tealdeer` instead, a Rust-based `tldr` implementation that provides the same `tldr` binary.

This PR updates `compile-setup.sh` to request `tealdeer` specifically under the `apt-get|nala` package manager branch. Other package manager branches remain unchanged as they distribute the package directly under `tldr`.

### Testing
Verified on **Debian Trixie**:

```bash
$ ./compile-setup.sh
...
tealdeer is already the newest version (1.7.2-1+deb13u1).
...
$ tldr nano
Text editor. An enhanced `pico` clone.
...

```

The `tldr` command resolves and functions as expected post-installation.

*Note: As with standard `tealdeer` behavior, a fresh installation requires running `tldr --update` once to populate the local cache.*

## Issues / other PRs related

N/A — Discovered while executing `compile-setup.sh` locally; no prior issue reported.

---

*This is my first open-source contribution! Any feedback or pointers regarding PR process standards are greatly appreciated. Thanks for reviewing!*


Abs313a:
@karsterr Thanks for addressing the Debian Trixie failure..

Could you select the package based on repository availability instead of replacing `tldr` for every `apt-get`/`nala` system?

The current branch also covers Ubuntu 22.04 LTS, where `tldr` is available but `tealdeer` is not, so this change would make the entire dependency installation fail under `sh -e`.

I believe the robust approach would be to run `apt update`, use `tldr` when available, and fall back to `tealdeer` otherwise. Could you also verify both Debian Trixie and Ubuntu 22.04?

